PR #3, opened 2023-12-15
Added basic DIV operator

andresnowak:
Added basic DIV operator and its tests for forward and backward.

**Extra:**
- An Idea: Maybe we could add to the NODE struct or (maybe another struct, I don't know how you would want to implement it) operator overloading to add the ability to be able to do `input / weights` instead of `DIV.forward(inputs, weights)`
- Also maybe in the *future*, if its possible, maybe it would be better to also have a graph of the forward operations, instead of doing them the moment the function is called (like `input / weight` doesn't do the division but instead adds the division operation to the forward graph and then we execute it later when necessary), to be able to do fusion operations and to be able to not have to create new memory of a tensor for each result (so to also be able to have lazy initialization of memory)

### DIFF
--- a/dainemo/autograd/ops/basics.mojo
+++ b/dainemo/autograd/ops/basics.mojo
@@ -120,7 +120,7 @@ struct DIV:
         """Backward operation of element wise division."""
         alias nelts: Int = simdwidthof[dtype]()
         # d(x/y) / dx = 1/y
-        # d(x/y) / dy = x/y^2
+        # d(x/y) / dy = -x/y^2
         if tensor_id == 0:
             let n2 = GRAPH.graph[GRAPH.get_node_idx(tensor_vec[1])]
             let res = elwise_op[dtype, nelts, div](1.0, n2.tensor)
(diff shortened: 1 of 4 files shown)


StijnWoestenborghs:
Great, thanks. The backward operation is missing a negative sign here:
d(x/y) / dy = -  x / y^2

StijnWoestenborghs:
- I agree in overloading the operators in Node in the near future. That would clean up the code base quit a lot and is probably quite an easy thing to do. 
- Will need to wrap my head around a good strategy on that forward graph indeed. Not sure how one could create a graph and fuse operators automatically with the tools available right now. And then there is big question to do that at runtime vs. at compiletime. What could work here is to implement fusion operators manually as a new operator. But probably this is part of a bigger discussion (and future headaches)

dainemo/autograd/ops/basics.mojo
```diff
@@ -92,7 +92,45 @@ struct MUL:
...
+        return GRAPH.create_graph_node[Self.backward](res, n1, Node[dtype](a_tensor))
+
+    @staticmethod
+    fn backward(
+        ug: Tensor[dtype], tensor_vec: DynamicVector[String], tensor_id: Int
+    ) -> Tensor[dtype]:
+        """Backward operation of element wise division."""
+        alias nelts: Int = simdwidthof[dtype]()
+        # d(x/y) / dx = 1/y
+        # d(x/y) / dy = x/y^2
```
StijnWoestenborghs, line 123: # d(x/y) / dy = - x/y^2

andresnowak:
Fixed the div backward operation.
- And yeah that's why I said that maybe in the future it would be possible to do fuse op, because right now I also think it wouldn't be possible, just for maybe simple operations like pure elementwise operations that happen in order in the graph. But I said this Idea, because if I remember correctly, modular mentioned that they they designed the engine so it can decide when to merge the operations, and that they do it manipulating the MLIR or something like that, butt they said that for now in the SDK available this wouldn't be possible, but that in the future they would add it.

StijnWoestenborghs (approved):
Great!

PR #5, opened 2023-12-19
Refactored get_broadcasting_shape

andresnowak:
### Refactored get_broadcasting_shape

From what I understand the function was already prepared to do a broadcast of more than a rank 2, so this commit only adds the ability to check the biggest rank of all tensors and create the bc_shape and none_bc of that size.

### Fixed a bug with the parser in basics.mojo

I also fixed a bug where import from `dainemo.utils.tensorutils import transpose_2D` inside the *elif* in *SUM* would cause the language server to crash, and I also removed the import from *DOT* struct, because after doing this change then the *test_regression* would now crash. So it seems there is a bug when importing in a local namespace?

## Extra
- I also think it is possible to remove the result tensor argument from the get_broadcasting_function no? because it isn't used.
- And I also see that the function isn't used in any operation, was it because the get_broadcasting_shape was only working for rank 2, or was there another reason?

### DIFF
--- a/dainemo/autograd/graph.mojo
+++ b/dainemo/autograd/graph.mojo
@@ -6,30 +6,31 @@ from dainemo.utils.tensorutils import elwise_op, zero
 from math import add, max
 
 
-
 struct Graph[dtype: DType = DType.float32, tracking: Bool = True](Stringable):
-    '''
+    """
     Keeps track of all the nodes and its relations in the computational graph.
-    Created during the forward pass and used by the backpard pass to 
+    Created during the forward pass and used by the backpard pass to
     compute gradients through autodiff.
-    '''
+    """
 
     var keys: DynamicVector[String]
     var graph: DynamicVector[Node[dtype]]
-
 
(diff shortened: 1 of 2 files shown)


StijnWoestenborghs:
I think we need a different approach on calculating the broadcasting shape.
Was looking into numpy and they are actually broadcasting empty arrays of the input shapes.
https://github.com/numpy/numpy/blob/v1.26.0/numpy/lib/stride_tricks.py#L434

This is probably connected to:
- Implementing a proper broadcasting function: https://numpy.org/doc/stable/user/basics.broadcasting.html
- Replacing and Getting rid of "batch_tensor_elwise_op"

StijnWoestenborghs:
I'm thinking now to just merge this and have proper broadcasting as one of the next priorities

StijnWoestenborghs (approved):

andresnowak:
Yeah, probably it would be better to have a correct broadcasting implementation. I will try and read the numpy implementation and maybe try to implement it.

PR #6, opened 2023-12-26
Added vectorization to transpose_2d and created a transpose function for any rank with vectorization

andresnowak:
### Added vectorization for transpose_2d
I added *vectorization* for this function using *simd_strided_store* and *simd_load*

### Added a transpose function for any rank
This PR also add a transpose function that works for any rank and has *parallelization* and *vectorization*

### Added new tests for transpose
Also this PR added new tests for the transpose function to test for more rank sizes. And I also divided the test to another file called *test_tensorutils_data*, this file generates the data to be used to test with the assert equals, I did this division to not add a lot of code to the original test file, so as to not make the *test_tensorutils* file more difficult to read, but also to not have to deal with the *mojo formater*, if i had an array that was bigger than what could be in the screen, the *mojo format* would make put each value of the array in a new line, so I dint want to add *108 lines* just to have the expected arrays.

### DIFF
--- a/dainemo/utils/tensorutils.mojo
+++ b/dainemo/utils/tensorutils.mojo
@@ -252,90 +252,76 @@ fn transpose_2D[dtype: DType, nelts: Int](t: Tensor[dtype]) -> Tensor[dtype]:
 
 
 @always_inline
+fn calculate_strides(shape: TensorShape) -> DynamicVector[Int]:
+    var strides = DynamicVector[Int](shape.rank())
+    strides.resize(shape.rank(), 1)
+
+    for i in range(shape.rank() - 2, -1, -1):
+        strides[i] = strides[i + 1] * shape[i + 1]
+
+    return strides
+
+
+@always_inline
 fn transpose[
     dtype: DType, nelts: Int
 ](t: Tensor[dtype], dim_0: Int, dim_1: Int) -> Tensor[dtype]:
     """
-    Create a new transposed tensor of the given tensor t.
+    Create a new tensor transposing dim_0 and dim_1.
     """
-
-    # Get the new shape of the transposed tensor
-    var new_shape = DynamicVector[Int](t.rank())
+    var axes = DynamicVector[Int](t.rank())
 
     for i in range(t.rank()):
         if i == dim_0:
-            new_shape.push_back(t.dim(dim_1))
+            axes.push_back(dim_1)
         elif i == dim_1:
-            new_shape.push_back(t.dim(dim_0))
+            axes.push_back(dim_0)
         else:
-            new_shape.push_back(t.dim(i))
+            axes.push_back(i)
 
-    var t_new = Tensor[dtype](new_shape)
+    return transpose[dtype, nelts](t, axes)
 
-    # Get the strides of the old and new tensors
-    var dims = StaticIntTuple[2]((min(dim_0, dim_1), max(dim_0, dim_1))) # last and first dimension (reading from right to left)
 
-    # example: tensor(2x3), strides (6, 3, 1)
-    var strides_old = DynamicVector[Int](t.rank() + 1)
-    var strides_new = DynamicVector[Int](t_new.rank() + 1)
-    strides_old.resize(t.rank() + 1, 1)
-    strides_new.resize(t_new.rank() + 1, 1)
+@always_inline
+fn transpose[dtype: DType, nelts: Int](t: Tensor[dtype]) -> Tensor[dtype]:
+    """
+    Create a new transposed tensor of the given tensor t.
+    """
+    var axes = DynamicVector[Int](t.rank())
 
     for i in range(t.rank() - 1, -1, -1):
-        strides_old[i] = strides_old[i + 1] * t.dim(i)
-        strides_new[i] = strides_new[i + 1] * t_new.dim(i)
+        axes.push_back(i)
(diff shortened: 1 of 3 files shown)


StijnWoestenborghs:
Now that is one amazing PR! 

I want to phase out the transpose_2D function and use this to fix the backwards DOT operator to work for any rank. (As I'll need it in the conv2d layer and in the mnist example.) There are currently 2 places where transpose_2D is used. Both in the backward operators:
- Backwards SUM, axis=1. But this is a workaround in itself and will not need transpose, but broadcasting to be implemented. So ignore this one
- Backwards DOT: Where transpose needs to **reverse the order of the axis**. (e.g. (4, 1, 28, 28) --> (28, 28, 1, 4)) As far as I understand a small API change to the transpose function would do it. Having a `new_shape: TensorShape` argument (+redoing the strides), would allow general axis reordering instead of swapping two only. 

And then possibly a function overload of the transpose function can implement the reverse order of the axis as default behaviour. This should also be default behaviour of [numpy's transpose.](https://numpy.org/doc/stable/reference/generated/numpy.transpose.html)

Is this a small change or does it mess with the indexing during parallelization and vectorization?

andresnowak:
The change would mess up the indexing. To make a transpose for any number of axis I have two ideas.

- one would be to have three fors. The first one would be the iteration of all the values in the tensor minus the last dimension, then the second for would iterate over all the dimensions a te sor has to do the operation to get index we are in each dimension and then do the operation to know where we are in the new tensor, and the third for would the last dimension using vectorization. Or something like this

- the other one would be using a vector to keep track of where we are in each dimension. But because we are using parallelization and least this way wouldn't be possible.

For the first idea maybe at least it would be possible to do an unroll of 2 to make it a little bit more efficient. But maybe there is better and more efficient way to do this implementation.

StijnWoestenborghs:
I spend some time on this today. Have a look at this strategy. It is a working python implementation for now:


```
import numpy as np

# Creating an array of zeros with shape (2, 3, 2)
arr = np.zeros((2, 3, 2))

value = 1
# Looping through each element and adding +1
for i in range(arr.shape[0]):
    for j in range(arr.shape[1]):
        for k in range(arr.shape[2]):
            arr[i, j, k] += value
            value += 1





def calculate_strides(shape):
    """Given the shape of a tensor, calculate its strides."""
    strides = [1]
    for size in reversed(shape):
        strides.insert(0, strides[0] * size)
    return strides[1:]


def transpose(tensor, axes):
    """
    Transpose the given tensor using the specified axes.
    """
    assert len(axes) == len(tensor.shape), "Axes don't match the tensor's dimensions."

    new_shape = [tensor.shape[axis] for axis in axes]  # Determine the shape of the new tensor
    new_tensor = np.zeros(new_shape)  # Initialize the new tensor

    # Calculate strides for the original and transposed shapes
    original_strides = calculate_strides(tensor.shape)
    transposed_strides = calculate_strides(new_shape)

    print("original_strides: ", original_strides, "transposed_strides: ", transposed_strides)


    # Map each element from the original tensor to the new, transposed tensor
    for i in range(tensor.size):

        # Step 1: Unravel the original linear index to multi-dimensional indices
        linear_index = i
        unraveled_indices = [0] * len(tensor.shape)
        for j in range(len(tensor.shape)):
            stride = original_strides[j]
            unraveled_indices[j], linear_index = divmod(linear_index, stride)
        
        # Step 2: Rearrange indices according to the transposition axes
        transposed_indices = [unraveled_indices[axis] for axis in axes]
        
        # Step 3: Ravel the indices back to a linear index using transposed strides
        new_linear_index = 0
        for j, index in enumerate(transposed_indices):
            new_linear_index += index * transposed_strides[j]

        # Move the element from the original tensor to the new, transposed tensor
        new_tensor.ravel()[new_linear_index] = tensor.ravel()[i]

    return new_tensor



# Example of usage:
axes = [2, 1, 0] 
transposed_tensor = transpose(arr, axes)
print(transposed_tensor)

print("----True-----")
print(np.transpose(arr, axes))
```

StijnWoestenborghs:
That'll be great, no rush. Planning to focus on performance afterwards, but first: make it work 😄. In the mean time I'm doing conv2D now.

andresnowak:
Oh with the Conv2d I also had some ideas, the only problem for me was thinking about how to save the padding and strides values to use in the backward pass I was thinking maybe using traits or having a dict in node that has extra params. But maybe you shave solved that already, just throwing some ideas.

StijnWoestenborghs:
Yeah, I had that problem with the SUM operator already. What you can do is save it as a parameter during forward pass, since you are saving the whole backward_fn as a parameter during forward pass. Line 268 in basics ops: `return GRAPH.create_graph_node[Self.backward[axis=axis]](res, n)`
If you have ideas already, drop them in the other PR! I just started.

andresnowak:
I changed the transpose function to work for any rank size. The only thing I don't like is that I had to use DynamicVector instead of a VariadicList because VariadicLists can't be modified, so to use the overloaded functions I had to use DynamicVector instead of VariadicList.

StijnWoestenborghs:
Amazing, wouldn't know any another way then DynamicVector indeed when choosing to do it this way. The comment is good to keep it in mind. 👌

StijnWoestenborghs (approved):

PR #11, opened 2023-12-30
Add CONV2D

StijnWoestenborghs:
Implements the conv2d operator and conv2d layer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ test:
 	mojo run -I . test/test_backward.mojo
 	mojo run -I . test/test_loss.mojo
 	mojo run -I . test/test_regression.mojo
-
-trait:
 	mojo run -I . test/test_traits.mojo
+
+conv:
+	. .venv/bin/activate && mojo run -I . test/test_conv.mojo
--- a/dainemo/autograd/graph.mojo
+++ b/dainemo/autograd/graph.mojo
@@ -172,6 +172,13 @@ struct Graph[dtype: DType = DType.float32, tracking: Bool = True](Stringable):
         self.keys = param_keys
         self.graph = param_graph
 
+    fn reset_all(inout self):
+        """
+        Resets the graph.
+        """
+        self.keys = DynamicVector[String]()
+        self.graph = DynamicVector[Node[dtype]]()
+
     fn reset_visited(inout self):
         """
         Marks visited as False for every Node in the graph.
--- a/dainemo/autograd/ops/conv.mojo
+++ b/dainemo/autograd/ops/conv.mojo
@@ -44,6 +44,7 @@ struct CONV2D:
             output.shape     [batch, out_channels, X, Y].
         """
         # TODO: Add bias
+        # TODO: calculate kernel_index and input_index using precalculated strides
 
         alias nelts: Int = simdwidthof[dtype]()
 
@@ -118,10 +119,49 @@ struct CONV2D:
     fn backward(
         ug: Tensor[dtype], tensor_vec: DynamicVector[String], tensor_id: Int
     ) -> Tensor[dtype]:
-        """Backward operation of 2D convolution."""
-        # TODO
-        return Tensor[dtype]()
+        """
+        Backward operation of 2D convolution.
+            
+        Upper gradient of shape: [batch, out_channels, X, Y].
+        """
+        
+        alias nelts: Int = simdwidthof[dtype]()
 
+        if tensor_id == 0:
+            # Inputs
+            # TODO
+
+            return Tensor[dtype]()
+        
+        elif tensor_id == 1:
+            # Kernel
+            # TODO
+
+            return Tensor[dtype]()
+
+        else: 
+            # Bias
+            # Sum of upper gradient over batch and X, Y dimensions
+            # out_channels == ug.dim(1) == bias.dim(0)
+            # TODO: calculate ug_index using precalculated strides
+            var result = Tensor[dtype](ug.dim(1))
+
+            for out_ch in range(ug.dim(1)):
+                var sum: SIMD[dtype, 1] = 0
+                for batch in range(ug.dim(0)):
+                    for x in range(ug.dim(2)):
+                        for y in range(ug.dim(3)):
+                            let ug_index = (
+                                batch * (ug.dim(1) * ug.dim(2) * ug.dim(3))
+                                + out_ch * (ug.dim(2) * ug.dim(3))
+                                + x * ug.dim(3)
+                                + y
+                            )
+                            sum += ug[ug_index]
+
+                result[out_ch] = sum
+
+            return result
 
 
 # <------------CONV3D------------>
--- a/dainemo/autograd/ops/conv.mojo
+++ b/dainemo/autograd/ops/conv.mojo
@@ -139,14 +139,14 @@ struct CONV2D:
             
             for batch in range(inputs.dim(0)):
                 for in_ch in range(inputs.dim(1)):
-                    for x in range(inputs.dim(2)):
-                        for y in range(inputs.dim(3)):
+                    for ix in range(inputs.dim(2)):
+                        for iy in range(inputs.dim(3)):
                             var result: SIMD[dtype, 1] = 0
                             for out_ch in range(ug.dim(1)):
                                 for kx in range(kernel.dim(2)):
                                     for ky in range(kernel.dim(3)):
-                                        let ux = x * stride - kx + padding
-                                        let uy = y * stride - ky + padding
+                                        let ux = ix * stride - kx + padding
+                                        let uy = iy * stride - ky + padding
 
                                         if ux < 0 or uy < 0 or ux >= ug.dim(2) or uy >= ug.dim(3):
                                             continue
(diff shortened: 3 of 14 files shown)


andresnowak:
## This is a first implementation idea I am doing based on what you have implemented:
But I don't know if you wanted to make the implementation simpler to read using instead slices.
```py
struct CONV2D:
    @staticmethod
    fn forward[
        padding: Int, stride: Int
    ](
        inputs: Tensor[dtype], kernel: Tensor[dtype], bias: Tensor[dtype]
    ):  # -> Node[dtype]:
        """
        Performs a 2D convolution on the input tensor using the kernel and bias.
        """
        # inputs.shape should be len 3 with [batch, X, Y]
        alias nelts: Int = simdwidthof[dtype]()

        let result_shape = get_result_shape[padding, stride](
            inputs.shape(), kernel.shape()
        )
        var outputs = Tensor[dtype](inputs.dim(0), result_shape[0], result_shape[1])
        let padded_inputs = pad[padding](inputs)

        var index_i = 0
        var index_j = 0
        let kernel_x_dim = kernel.shape()[-2]
        let kernel_y_dim = kernel.shape()[-1]
        for batch in range(outputs.dim(0)):
            for i in range(outputs.dim(1)):
                for j in range(outputs.dim(2)):
                    # Get fragment of padded inputs
                    let slice_x = slice(index_i, index_i + kernel_x_dim)
                    let slice_y = slice(index_j, index_j + kernel_y_dim)
                    # let fragment = tslice[dtype, nelts](padded_inputs, slice(0, inputs.dim(0)), slice_x, slice_y)

                    # TODO
                    # output = np.sum((fragment*kernel.data), axis=(1,2)) + bias.data
                    # outputs[:, i, j] = output

                    # Iterate over kernel and multiply with fragment
                    var result: SIMD[DType.float32, 1] = 0
                    for k in range(kernel_x_dim):
                        for l in range(kernel_y_dim):
                            let padded_index = (
                                batch * padded_inputs.dim(1) * padded_inputs.dim(2)
                                + (index_i + k) * padded_inputs.dim(2)
                                + (index_j + l)
                            )
                            let kernel_index = (k * kernel.dim(1) + l)
                            result += padded_inputs[padded_index] * kernel[kernel_index]

                    let output_index = (
                        batch * outputs.dim(1) * outputs.dim(2) + i * outputs.dim(2) + j
                    )
                    outputs[output_index] = result

                    # Increment index by stride
                    index_j += stride
                index_j = 0
                index_i += stride
            index_i = 0
```
But I feel that creating a new tensor with is unnecessary. Because padding is only one value in reality, i feel the tensor can be iterated in a way where we account for this so as to not use more memory and create a copy of the tensor, but because right now we don't care about performance maybe we could leave it for now.

## Extra
- The other things is I don't know if you would want to make conv2d the same as how in pytorch work, because right now this implementation is different, because we only accept mini_batch and the height and width, but the pytorch one also accepts amount of input and output channels

StijnWoestenborghs:
Nice, I actually like this implementation more then what I was about to do with slices. As it allows for more optimisations (like the padding.) The reason I had the get_result_shape, pad & unpad separated was because they are common function between conv2d, conv3d, maxpool2d & maxpool3d. Let me see how I can write a test for this already.

Thanks for pointing out the in/out_channels. That is definitely something I was missing and that needs to be fixed.

StijnWoestenborghs:
I'm thinking we'll need slicing anyway in the backwards unpad. (for now, without the extra optimizations.) 
So I'll keep working on it

StijnWoestenborghs:
@andresnowak Eventually removed padding and slicing and added in-, out-channels. All in one go now with room for parallelizatio/vectorization. However the test is still indicating a mistake somewhere.

andresnowak:
It seems the problem was in this lines
```py
let input_index = (
  batch * (inputs.dim(1) * inputs.dim(2) * inputs.dim(3)) 
  + in_ch * (inputs.dim(2) * inputs.dim(3)) 
  + ix * inputs.dim(2) 
  + iy
 )
let kernel_index = (
    out_ch * (inputs.dim(1) * kernel.dim(2) * kernel.dim(3)) 
    + in_ch * (kernel.dim(2) * kernel.dim(3)) 
    + kx * kernel.dim(2) 
    + ky
)
```
in `ix * inputs.dim(2)` and `kx * kernel.dim(2)` it should be `ix * inputs.dim(3)` and `kx * kernel.dim(3)`


And right now I have the function like this
```py
struct CONV2D:
    @staticmethod
    fn forward[
        padding: Int, stride: Int, padding_mode: Int = 0
    ](inputs: Tensor[dtype], kernel: Tensor[dtype], bias: Tensor[dtype]) -> Node[dtype]:
        """
        Performs a 2D convolution on the input tensor using the kernel and bias.
            inputs.shape     [batch, in_channels, X, Y]
            kernel.shape     [out_channels, in_channels, X, Y] (or weights)
            bias.shape       [out_channels].
            output.shape     [batch, out_channels, X, Y].
        """
        # TODO: Add bias

        alias nelts: Int = simdwidthof[dtype]()

        let result_shape = get_result_shape[padding, stride](
            inputs.shape(), kernel.shape()
        )

        var outputs = Tensor[dtype](
            inputs.dim(0), kernel.dim(0), result_shape[0], result_shape[1]
        )

        @parameter
        fn kernel_iteration(
            batch: Int, in_ch: Int, out_ch: Int, x: Int, y: Int
        ) -> SIMD[DType.float32, 1]:
            var result: SIMD[DType.float32, 1] = 0
            for kx in range(kernel.dim(2)):
                for ky in range(kernel.dim(3)):
                    let ix = x * stride - padding + kx
                    let iy = y * stride - padding + ky
                    let kernel_index = (
                        out_ch * (kernel.dim(1) * kernel.dim(2) * kernel.dim(3))
                        + in_ch * (kernel.dim(2) * kernel.dim(3))
                        + kx * kernel.dim(3)
                        + ky
                    )
                    if ix < 0 or iy < 0 or ix >= inputs.dim(2) or iy >= inputs.dim(3):
                        result += padding_mode * kernel[kernel_index]
                        continue

                    let input_index = (
                        batch * (inputs.dim(1) * inputs.dim(2) * inputs.dim(3))
                        + in_ch * (inputs.dim(2) * inputs.dim(3))
                        + ix * inputs.dim(3)
                        + iy
                    )

                    result += inputs[input_index] * kernel[kernel_index]

            return result

        for batch in range(inputs.dim(0)):
            for out_ch in range(outputs.dim(1)):
                for x in range(outputs.dim(2)):
                    for y in range(outputs.dim(3)):
                        var result: SIMD[DType.float32, 1] = 0
                        for in_ch in range(inputs.dim(1)):
                            result += kernel_iteration(batch, in_ch, out_ch, x, y)

                        let output_index = (
                            batch * (outputs.dim(1) * outputs.dim(2) * outputs.dim(3))
                            + out_ch * (outputs.dim(2) * outputs.dim(3))
                            + x * outputs.dim(3)
                            + y
                        )
                        outputs[output_index] = result

        return GRAPH.create_graph_node[Self.backward](outputs, inputs, kernel, bias)
```

I divided it into a f unction to remove the identation, I feel it is a little bit more readable, but maybe I am wrong.

Will try and see about removing the if for the padding and also see about vectorization and adding the bias.

StijnWoestenborghs:
Amazing, I'm doing backward op now.

About vectorization: (I got the idea from here https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/conv.cpp#L193-L240)
The check `if ix < 0 or iy < 0 or ix >= inputs.dim(2) or iy >= inputs.dim(3):` Is actually only required for the edges in the X and Y dimension of the outputs matrix. Means we can split up this part ... into 5 cases and only check the bounds where needed.
```
                for x in range(outputs.dim(2)):
                    for y in range(outputs.dim(3)):
```
Check for the edges and for the base case where there is no risk of getting out of bounds. The if statement is not needed.
So we could create 2 functions `kernel_iteration_all_check` and `kernel_iteration_no_check` and have much better vectorization for that second function (which contains the majority of the conv operation.

The whole structure would look something like:
```
         for batch in range(inputs.dim(0)):
            for out_ch in range(outputs.dim(1)):

                # Case 1: TOP       x might put us out of bounds (check)
                for x in range(0, border_X1):
                    for y in range(outputs.dim(3)):
                        ...
                        kernel_iteration_all_checks(...)
                        ...

                for x in range(border_X1, border_X2):
                    # Case 2: LEFT      y might put us out of bounds (check)
                    for y in range(0, border_Y1):
                        ...
                        kernel_iteration_all_checks(...)
                        ...

                    # Case 3: Base case, in bounds (no checking needed, in the majority of cases)
                    for y in range(border_Y1, border_Y2):
                        ...
                        kernel_iteration_no_checks(...)
                        ...

                    # Case 4: RIGHT     y might put us out of bounds (check)
                    for y in range(border_Y2, outputs.dim(3)):
                        ...
                        kernel_iteration_all_checks(...)
                        ...
                
                # Case 5: BOTTOM    x might put us out of bounds (check)
                for x in range(border_X2, outputs.dim(2)):
                    for y in range(outputs.dim(3)):
                        ...
                        kernel_iteration_all_checks(...)
                        ...
```

It adds quite some volume to the operation. And we don't have to do it now (as performance is not the highest priority atm).
But good to take into account when you are looking at vectorization.

StijnWoestenborghs:
@andresnowak  The conv2d operator is functionally working now. Tested both forward and backward methods against the `torch.nn.functional.conv2d`. 
I'm thinking of closing this PR now, so it's easier for you as well.  Feel free to open a new PR if you are working looking at vectorization. Really appreciate it

PR #12, opened 2024-01-07
maxpool2d

StijnWoestenborghs:
**Adding maxpool2d operator and layer.**

Implements:
- maxpool2d forward and backward operator.
- Support for (h, w) values in padding, stride and dilation.
- Using the precalculated strides from Node.
- MaxPool2d nn layer.

Note that in_channels of the layer is a parameter since kernel_shape is required to be a parameter as well.
And I couldn't figure out how to check the (padding <= kernel_size / 2) at compile time. And at the same time, avoid that the whole MaxPool2d needs to be tagged as raises. I'm not exactly sure how expensive that is performance wise. That is why I mainly try to avoid to use raises everywhere.


dainemo/autograd/ops/pool.mojo
```diff
@@ -0,0 +1,146 @@
...
+        var outputs = Tensor[dtype](
+            inputs.tensor.dim(0), kernel_shape[0], result_shape[0], result_shape[1]
+        )
+        let outputs_strides = calculate_strides(outputs.shape())
+
+        for batch in range(inputs.tensor.dim(0)):
+            for in_ch in range(inputs.tensor.dim(1)):
+                for x in range(outputs.dim(2)):
+                    for y in range(outputs.dim(3)):
+                        var max_val: SIMD[dtype, 1] = -1e9
```
andresnowak, line 42: I think this could be changed with math.limit.min_finite or math.limit.neginf in the forward and backward pass, to be able to also represent the biggest value for any dtype. And also why -1e9, isn't the biggest value for 32bits (approx 1.1 x 10^-38), or are we never going to have a smaller value than -1e9 in maxpool?

dainemo/autograd/ops/pool.mojo
```diff
@@ -0,0 +1,146 @@
...
+                                    max_val = val
+
+                        let out_idx = (
+                            batch * outputs_strides[0]
+                            + in_ch * outputs_strides[1]
+                            + x * outputs_strides[2]
+                            + y
+                        )
+
+                        outputs[out_idx] = max_val
```
andresnowak, line 76: Extra: And If you want I think this could be changed to be implemented in the same way as the conv2d function, to be able to remove the if inside the fors using the border calculations

dainemo/autograd/ops/pool.mojo
StijnWoestenborghs: Interesting, didn't know about these. Will fix it

dainemo/autograd/ops/pool.mojo
StijnWoestenborghs: Yes, and probably we can also avoid those if's in the backward pass for both maxpooling and conv2d. Was thinking to implement this later indeed (after MNIST and when looking at vectorization/parallelization). But very nice you figured out the borders in the conv2d forward pass, so we have an example on how to do it.

StijnWoestenborghs: If good for you, can be merged

andresnowak (approved):

PR #13, opened 2024-01-10
Conv2D

andresnowak:
## This Pull request adds
- Support for different W and H values for padding in Conv2D [**Finished**]
- Edited the get_result_shape function to use // instead of floor and to int, just to make the function a little bit shorter and cleaner (both operations should give the same result) [**Finished**]
- Support for different W and H values for strides in Conv2D [**Finished**]
- Added more tests for Conv2D backward and forward pass [**Finished**]
- Added more tests for get_result_shape for Conv [**Finished**]
- Optimization that removed ifs in Conv2d Forward pass using the MLX Conv2D as an example [**Finished**]
- Added strides for the Tensor (At least while we wait for tensors to have it natively) in Node [**Finished**]
- Changed the operations in Conv2D to now use pre-calculated strides [**Finished**]
- Fixed a bug in the calculation of ix and iy in backward pass of Conv2D for kernel [**Finished**]
- Fixed a bug in the calculation of ux and uy in backward pass of Conv2D for inputs [**Finished?**]
	- The problem for this operations is that the calculation using stride wasn't really considering the stride, sometimes an ix or iy value would act on less ux and uy values than other ix and iy or even not contribute to any ux or uy
	- I fixed this bug by changing how we obtain the positions, instead of getting ux and uy I changed it to instead get ix and iy, I wanted to fix the calculation of ux and uy but I haven't found a way in how to calculate the kx_start and ky_start values. This was the version that I was trying (and kx and ky start is necessary because the formula for ux and uy would give ""*correct results*"" when giving it kx and ky values that the ix and iy values would never touch):
	- ```py
	   var res = Tensor[dtype](inputs.shape())

            for batch in range(inputs.dim(0)):
                for in_ch in range(inputs.dim(1)):
                    for ix in range(inputs.dim(2)):
                        for iy in range(inputs.dim(3)):
                            var result: SIMD[dtype, 1] = 0
                            # fix kx_start and ky_start values
                            let kx_start = (
                                ((ix + padding[0]) * stride[1]) % kernel.dim(2)
                            )
                            let ky_start = (
                                ((iy + padding[1]) * stride[0]) % kernel.dim(3)
                            )
                            for out_ch in range(ug.dim(1)):
                                for kx in range(kx_start, kernel.dim(2), stride[0]):
                                    for ky in range(ky_start, kernel.dim(3), stride[1]):
                                        let ux = (ix - kx + padding[0]) // stride[0]
                                        let uy = (iy - ky + padding[1]) // stride[1]

                                        if (
                                            ux < 0
                                            or uy < 0
                                            or ux >= ug.dim(2)
                                            or uy >= ug.dim(3)
                                        ):
                                            continue

                                        let kernel_index = (
                                            out_ch * kernel_strides[0]
                                            + in_ch * kernel_strides[1]
                                            + kx * kernel_strides[2]
                                            + ky
                                        )

                                        let ug_index = (
                                            batch * ug_strides[0]
                                            + out_ch * ug_strides[1]
                                            + ux * ug_strides[2]
                                            + uy
                                        )

                                        result += kernel[kernel_index] * ug[ug_index]

                            let input_index = (
                                batch * inputs_strides[0]
                                + in_ch * inputs_strides[1]
                                + ix * inputs_strides[2]
                                + iy
                            )
                            res[input_index] = result

            return res```
### New
- Support for Dilation [**Finished**]
- Propagate changes to Conv2D Layer [**Finished**]

### DIFF
--- a/dainemo/autograd/node.mojo
+++ b/dainemo/autograd/node.mojo
@@ -3,21 +3,21 @@ from tensor import Tensor, TensorShape
 
 from dainemo import GRAPH
 from dainemo.utils.uuid import uuid
-from dainemo.utils.tensorutils import fill, elwise_op, tsum
+from dainemo.utils.tensorutils import fill, elwise_op, tsum, calculate_strides
 
 
-fn backward_fn_placeholder[dtype: DType](
-        ug: Tensor[dtype],
-        tensor_vec: DynamicVector[String],
-        tensor_id: Int
-    ) -> Tensor[dtype]:
+fn backward_fn_placeholder[
+    dtype: DType
+](ug: Tensor[dtype], tensor_vec: DynamicVector[String], tensor_id: Int) -> Tensor[
+    dtype
+]:
     print("[ERROR]: Backward function placeholder")
     return Tensor[dtype](ug.shape())
 
 
 @value
 struct Node[dtype: DType = DType.float32](CollectionElement, Stringable):
-    '''
+    """
     A Node data structure that is used to build the computational graph.
         - tensor: The tensor that is stored in the node.
         - requires_grad: If true, the node will be included in the backward pass.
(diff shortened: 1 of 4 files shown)


StijnWoestenborghs:
Very nice (again) 
- Interesting to see that if you put an integer (e.g. 1) in something that expects a StaticIntTuple[2] It creates (1, 1) automatically.
The change for supporting H, W in padding and stride has also consequences for the ... nn/layers/conv.mojo 
Could you propagate that change in the Conv2d layer as well? The only thing I'm not sure about is the API. 
```
struct Conv2d[
    padding: StaticIntTuple[2],
    stride: StaticIntTuple[2],
]:
```
using this or `Tuple[Int, Int]` to be more python-like ...
And same for the kernel_size int the init function. You would be able to avoid overloading when using StaticIntTuple[2] though.

- I like the strides to be added to Node for now. Should we file a feature request for strides native into Tensor? Could be a property that only requires change on their reshape method

- If you are looking at dilation. get result shape could be updated like this (that's the only thing looked at for now):
```
    let result_x_dim = (
        (input_shape[-2] + (2 * padding[0]) - dilation[0] * (kernel_shape[-2] - 1) - 1) // stride[0]
    ) + 1
    let result_y_dim = (
        (input_shape[-1] + (2 * padding[1]) - dilation[1] * (kernel_shape[-1] - 1) - 1) // stride[1]
    ) + 1
```

- I'm putting a graph revision on the backlog. Maybe we need to rethink the api of the Backward function:
it think those `GRAPH.graph[GRAPH.get_node_idx(tensor_vec[0])]` are way too expensive as it loops over all the nodes in the graph to find the node, especially for big graphs.
When lifetimes are supported, it can remove any copy of a Node element that is there right now. Then can have a look at this as well.

andresnowak:
1. the reason why I used StaticTuple and not tuple is because
	1. Because of the tuple api, Where you have to use get[Int, AnyRegType] and that it returns a capturing function instead of the value
	2. Because We could change the get_result_shape funciton to work for 1d, 2d and 3d, based on the size we pass for the staticTuple, and use a for to do the calculation for h, w, etc..
	3. But I don't understand why tuple and staticTuple are so different, and why staticTuple can't be automatically converted to a tuple
2. Yeah I feel we could open an issue in mojo so they add strides to tensors, even though I would think they already have plans to add it. Because I still don't understand why there exists tensor, ndbuffer and dynamicNdBuffer, and why NDbuffer has more features than tensor. But If you want, would you like to open the issue or would you want me to open it, but I don't know what I could write about for the issue.
4. And for the graph revision where you think about something like changing the graph api to make it like the one in MLX or tinygrad or something like that (but this would also change how the forward operations work). Or where you thinking about a smaller change, to just try to fix that specific problem.

StijnWoestenborghs:
1/2. I think it's more then fine to use StaticIntTuple right now. And let's indeed leave the strides feature request for what it is. Minor issue and I kind off like it being a property of Node now. 

3. About the **graph revision**: Probably the revision is performance wise, e.g. check how much can be done at compile-time already, and doing as much as possible there... For sure let's wait for Lifetimes first. As the graph right now is suffering from some unneeded copies because of this. As I see it, it could be a big change that changes the forward and backward API of the ops.

Is this ready to merge btw? 
Checked it again and it is looking good for me.

StijnWoestenborghs:
Here it is: https://github.com/modularml/mojo/issues/1624

StijnWoestenborghs (approved):

PR #15, opened 2024-01-16
[Changed] Implement Softmax

andresnowak:
This branch implements:
- Softmax [**FINISHED**]
	- The problem with this one is that the backward pass changes for i == j and i != j (so it is a little bit more complicated, and I still don't understand it a 100% to know how to implement it). [softmax derivative](https://shivammehta25.github.io/posts/deriving-categorical-cross-entropy-and-softmax/)
	- So I don't know if to try implement the backward pass, or just implement the forward using operations that already have the backward pass implemented. [In the end I think it is better to implement the forward using operation for now. And for the backward operations in the future, maybe we should change them to be like in mlx or Jax using the vjp and jvp]
	- **To be able to implement softmax** we need:
		- MAX
			- Implement tmax for any n axis [**FINISHED**]
			- Implement Max Forward [**FINISHED**]
			- Implement Max backward [**FINISHED**]
		- SUM
			- Refactor tsum to work for any n axis [**FINISHED**]
			- Refactor sum backward to work for any n axis [**FINISHED**]
		- **So to be able to implement this it seems that maybe we have to refactor the broadcast and batch_elementwise function. So maybe we should refactor the broadcasting**
- logSoftmax [**WIP**]
- Implement broadcasting [**FINISHED**]
- Implement broadcast_elwise [**FINISHED**]

*I can't change the branch name because that would close the pull request, so I will leave that name for now.*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,5 @@ test:
 	mojo run -I . test/test_loss.mojo
 	mojo run -I . test/test_regression.mojo
 	mojo run -I . test/test_traits.mojo
-
-conv:
 	. .venv/bin/activate && mojo run -I . test/test_conv.mojo
+	. .venv/bin/activate && mojo run -I . test/test_pool.mojo
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ test:
 	mojo run -I . test/test_backward.mojo
 	mojo run -I . test/test_loss.mojo
 	mojo run -I . test/test_regression.mojo
-
-trait:
 	mojo run -I . test/test_traits.mojo
+
+conv:
+	. .venv/bin/activate && mojo run -I . test/test_conv.mojo
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 | Task       | Dataset | Forward | Backward | Training |
 |------------|---------|---------|----------|----------|
 | REGRESSION |   ✅    |   ✅    |    ✅    |    ✅    |
-| CONV       |   ✅    |   ❌    |    ❌    |    ❌    |
+| MNIST      |   ✅    |   ❌    |    ❌    |    ❌    |
 
 
 <sub>Regression example:</sub>
@@ -67,7 +67,7 @@ WIP: Work in progress
 | TRANSPOSE  |   ✅   |
 | FLATTEN    |   ✅   |
 | RESHAPE    |   ✅   |
-| CONV2D     |   ❌   |
+| CONV2D     |   ✅   |
 | CONV3D     |   ❌   |
 | MAXPOOL2D  |   ❌   |
 | MAXPOOL3D  |   ❌   |
@@ -104,7 +104,7 @@ WIP: Work in progress
 | SEQUENTIAL |   ❌   |
 | LINEAR     |   ✅   |
 | DROPOUT    |   ❌   |
-| CONV2D     |   ❌   |
+| CONV2D     |   ✅   |
 | CONV3D     |   ❌   |
 | MAXPOOL2D  |   ❌   |
 | MAXPOOL3D  |   ❌   |
@@ -146,4 +146,4 @@ This project is licensed under the Apache License 2.0 - see the [LICENSE](LICENS
 
 ## Acknowledgements
 
-A huge thanks to [pranftw](https://github.com/pranftw) for the inspiration and support! This project is inspired by [neograd](https://github.com/pranftw/neograd) which forms the basis of the implementation.
+A huge thanks to [pranftw](https://github.com/pranftw) for the inspiration and support! This project is inspired by [neograd](https://github.com/pranftw/neograd).
--- a/README.md
+++ b/README.md
@@ -63,7 +63,8 @@ WIP: Work in progress
 | EXP        |   ✅   |
 | LOG        |   ✅   |
 | POW        |   ✅   |
-| SUM        |   ✅*  |
+| SUM        |   ✅   |
+| MAX        |   ✅   |
 | TRANSPOSE  |   ✅   |
 | FLATTEN    |   ✅   |
 | RESHAPE    |   ✅   |
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ WIP: Work in progress
 | RELU      |   ✅   |
 | SIGMOID   |   ✅   |
 | TANH      |   ✅   |
-| SOFTMAX   |   ❌   |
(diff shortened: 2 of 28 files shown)


StijnWoestenborghs:
I agree on implementing proper broadcasting for this. Good to be aware to know what parts will be affected.
Connected to broadcasting, I can think of:
- tensorutils **tsum**, **batch_tensor_elwise_op**, (maybe **elwise_op** between tensor and scalar, but not sure as it might decrease performance) 
- graph **get_broadcasting_shape**
- node **unbroadcast_data**

I can have a look at this if you want. Or are you looking at it already?

andresnowak:
No I still haven't look into this, because I didn't know what problem would arise if I changed batch_elwise and sum and tsum, and I still don't understand what unbroadcast data and get_broadcasting_shape do exactly. So I would appreciate the help. I will also try and see about the numpy broadcast implementation

andresnowak:
This is now the part where I don't know what to do. Like isn't there another way to do this more efficiently.
```py
fn foo():
    let t = Tensor[DType.float32](4, 3, 2)
    for i in range(10):
        var index = 0
        for j in range(t.rank()): # do this for every tensor
            index += i % t.shape()[j] * t.stride()[j]
```
So like when doing broadcasting or doing reshapes without modifying the memory to make it contiguous. is this the only way to iterate the tensors?, and that is the reason ndbuffer wants a compile time shape to do this efficiently (to be able to unroll it?)

StijnWoestenborghs:
Had a look at MLX its sum. As far as I understand this is pseudo-code of their (cpu-eval) sum op: 
```
def sum(array, axes, keepdims):
      out_shape, sorted_axes = compute_reduce_shape(axes, array.shape(), keepdims)
      out == array(output_shape)
      reduction_op(array, out, sorted_axes, 0, vectorized_sum_f(...))
      return out
```
A very similar idea to your general reduce function there. With:
that reduction_op function here: https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/reduce.h#L255C1-L282C12
```
def reduction_op(in, out, axes_, init, vectorized_op_f(...))
 ...
     # GeneralContiguousReduce and ContiguousReduce
     # Unrolling the following loop (and implementing it in order for ContiguousReduce should hold extra performance boost.
     for (i = 0 .. out.size() )
        int offset = elem_to_loc(i, shape, strides);
        nd_loop(   # (i think a loop unrolling helper function)
            vectorized_op_f(... offset ....),
         )
```
And this elem_to_loc has similar functionality to what we are doing now:
https://github.com/ml-explore/mlx/blob/main/mlx/utils.cpp#L150C1-L161C2
With elem_to_loc equivalent to:
```
def elem_to_loc(elem, shape, strides):
    loc = 0
    for i in reversed(range(len(shape))):
        q, r = divmod(elem, shape[i])
        loc += r * strides[i]
        elem = q
    return loc
```
It is a total maze but by reading this I believe this is the right track. It looks like we should be able to unroll now. However, I believe you are right, the mojo unroll decorator only works when the loop bounds are compile time constants.

StijnWoestenborghs:
Alright, broadcasting, reduce, N dim MAX and SUM, ... Everything should be there for Softmax now. 
I merged main as wel as there were some conflicts.

andresnowak:
I Added the softmax activation function as an activation, copying the idea of [mlx](https://github.com/ml-explore/mlx/blob/4a9012cba0c50383d3cdc6f39534a016141fec5f/python/mlx/nn/layers/activations.py#L44). Maybe this can be done another way, but here for the design we have now we could add all the activation functions here, and some of them won't have custom backward operator. I didn't do the custom backward operator of softmax because of what I mentioned before, I didn't know how to implement it with the current design (but maybe it is possible), so maybe in the future we should change the backward operator to work like in jax and mlx

StijnWoestenborghs:
I see multiple frameworks going for this way indeed. I mean, it will definitely work. Just wonder what to overhead is of having ~5 additional nodes in the graph. (And operator fusion as a solution pops up again as well.) I'm fine having it this way right now. 
Also because having the backward functions implemented manually is tons of extra work and the doing graph revision has higher priority then.

StijnWoestenborghs:
Nice I was doing the same. 3 very small things:
- Can you change the nn.RELU -> nn.ReLU, just to conform with the pytorch api
- Implement __call__ methods in the activations, so also in the tests nn.Softmax[0](x) is possible (just like nn.Linear)
- Update the readme with ✅

andresnowak:
Done, or would you want me to change inout self to self, to be able to use let instead of var. I used the inout self because that's how the loss functions are implemented, but it isn't neccessary no, we never modify the class.

StijnWoestenborghs:
Would be nice the be able to use let. is it calling `return Self.forward(input)` then and on staticmethod? 
Also take those extra in readme: I figure it was quite some effort :p 
| SUM        |   ✅   |
| MAX        |   ✅   |

andresnowak:
The thing is, I wasn't able to use call as a static method, it would give errors, so I changed to be as how linear or mse works, where you initialize the struct and then use it.

StijnWoestenborghs:
Ah, yeah as George would say. ReLU in a class makes no sense anyway haha
https://www.youtube.com/watch?v=YR-1fyQXK1o  (sec. 50)

andresnowak:
Yeah, so I'm not sure exactly how to leave the like the structs, if to just have a static_method for the forward, or if we leave it as an initialized struct, but just remove the inout from the function calls, to be able to use let. And the other thing is that doing it with initialized structs makes it so you have to initialize softmax and logsoftmax for different dims.

StijnWoestenborghs:
Yeah I think that is fine for now. Just to be as close to the mnist example.
- `@static_method` in forward, and `__call__(self, ...)` for let.
- and struct inits for softmax and logsoftmax
We can change the api later. But this is probably closest to pytorch for now... I'm not sure I like it 100% either
It will work here: https://github.com/StijnWoestenborghs/dainemo/blob/main/examples/mnist.py#L32C1-L52C45

andresnowak:
Done. 
If I understood correctly the struct inits for softmax and logsoftmax where refering to the compile time values no?, because we still need the __init__ function for all structs to be able to use the __call__ operator.

StijnWoestenborghs (approved):

StijnWoestenborghs:
Indeed, let's have a bigger look on the overall design of the API later I would say.
This was very good progress I would say. 🔥 (MNIST tomorrow)

PR #16, opened 2024-01-21
MNIST 

StijnWoestenborghs:
Implements the final pieces required for the MNIST benchmark to train:
- **Cross Entropy Loss** [FINISHED]
- <s>**Binary Cross Entropy Loss** (BCELoss)</s>
- <s>**nn.Sequential** Layer [WIP]</s>
- **mnist.mojo** example [WIP]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .venv
+__pycache__
 
 dainemo.📦
 
--- a/Makefile
+++ b/Makefile
@@ -41,3 +41,4 @@ test:
 	mojo run -I . test/test_mlops.mojo
 	. .venv/bin/activate && mojo run -I . test/test_conv.mojo
 	. .venv/bin/activate && mojo run -I . test/test_pool.mojo
+	. .venv/bin/activate && mojo run -I . test/test_models_torch.mojo
--- a/Makefile
+++ b/Makefile
@@ -36,5 +36,8 @@ test:
 	mojo run -I . test/test_loss.mojo
 	mojo run -I . test/test_regression.mojo
 	mojo run -I . test/test_traits.mojo
+	mojo run -I . test/test_activations.mojo
+	mojo run -I . test/test_broadcasting.mojo
+	mojo run -I . test/test_mlops.mojo
 	. .venv/bin/activate && mojo run -I . test/test_conv.mojo
 	. .venv/bin/activate && mojo run -I . test/test_pool.mojo
--- a/README.md
+++ b/README.md
@@ -78,8 +78,8 @@ WIP: Work in progress
 | Task      | Status |
 |-----------|--------|
 | MSE       |   ✅   |
-| CE        |   ❌   |
-| BCE       |   ❌   |
+| CE        |   ✅   |
+| BCE       |   ✅   |
 | SoftmaxCE |   ❌   |
 
 ### Activations
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 | Task       | Dataset | Forward | Backward | Training |
 |------------|---------|---------|----------|----------|
 | REGRESSION |   ✅    |   ✅    |    ✅    |    ✅    |
-| MNIST      |   ✅    |   ❌    |    ❌    |    ❌    |
+| MNIST      |   ✅    |   ✅    |    ❌    |    ❌    |
 
 
 <sub>Regression example:</sub>
--- a/dainemo/autograd/ops/basics.mojo
+++ b/dainemo/autograd/ops/basics.mojo
@@ -35,6 +35,14 @@ struct ADD:
         return GRAPH.create_graph_node[Self.backward](res, n1, n2)
 
     @staticmethod
+    fn forward(n1: Node[dtype], a: SIMD[dtype, 1]) -> Node[dtype]:
+        """Forward operation of tensor-scalar addition."""
+        let res: Tensor[dtype] = elwise_op[dtype, nelts, add](n1.tensor, a)
+        var a_tensor: Tensor[dtype] = Tensor[dtype](1)
+        a_tensor[0] = a
+        return GRAPH.create_graph_node[Self.backward](res, n1, Node[dtype](a_tensor))
+
+    @staticmethod
     fn backward(
         ug: Tensor[dtype], tensor_vec: DynamicVector[String], tensor_id: Int
     ) -> Tensor[dtype]:
--- a/dainemo/autograd/ops/basics.mojo
+++ b/dainemo/autograd/ops/basics.mojo
@@ -11,7 +11,6 @@ from dainemo.utils.tensorutils import (
     elwise_pow,
     elwise_transform,
     fill,
-    broadcast_elwise_op,
     transpose,
     calculate_strides,
 )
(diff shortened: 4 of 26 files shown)


andresnowak:
What do you think of this for sequential:
```py
trait Module:
    fn forward(self, inputs: Node[dtype]) -> Node[dtype]:
        pass

    fn __call__(self, inputs: Node[dtype]) -> Node[dtype]:
        pass
```

```py
struct Sequential:
    '''
    Abstract container of layers
    Sequentially passes the output of one layer the input of the next layer.
    '''

    var layers: VariadicListMem[Module]

    fn __init__(inout self, *layers: Module):
        self.layers = layers

    fn forward(self, inputs: Node[dtype]) -> Node[dtype]:
        var outputs: Node[dtype] = inputs
        for i in range(self.layers.__len__()):
            outputs = __get_address_as_lvalue(self.layers[i]).forward(outputs)
        return outputs

    fn __call__(self, inputs: Node[dtype]) -> Node[dtype]:
        return self.forward(inputs)
```
The only problem is that:
1. we can't have the forward staticmethods in int the activation functions.
2. The sequential could be a struct that receives all the modules at compile time to be able to unroll them, but:
	1. It seems it isn't possible to receive right now a variadic list as a compile time value (or maybe I did something wrong, but I would get a parsing crash running it)
	2. We can't also use StaticTuple because Module isn't a register passable value.
	3. But we could at least receive a compile time value of the amount of modules we have, to be able to unroll it at least

StijnWoestenborghs:
Yeah I was basically trying this, but I don't think it is possible yet. 

```
trait Module:
    fn __call__(self, inputs: Int) -> Int:
        ...

struct A(Module):
    fn __init__(inout self): pass

    fn __call__(self, inputs: Int) -> Int:
        print("A")
        return 1 + inputs

struct B(Module):
    fn __init__(inout self): pass

    fn __call__(self, inputs: Int) -> Int:
        print("B")
        return 2 + inputs

fn call_all(*all: Module):
    for s in all:
        print(__get_address_as_lvalue(s)(1))
        print()

fn main():
    var a = A()
    var b = B()
    call_all(a, b)     # --> invalid call to 'call_all': argument #0 cannot be converted from 'A' to 'Module'
```

I can't seem to find a way to have them inside a function call...

andresnowak:
Yeah I see I thought that it worked at runtime but now I see it also doesn't (I thought I tested it). It seems this is the only way to do it for now:
This is an example from here [example](https://github.com/rd4com/mojo-learning/blob/main/tutorials/traits.md)
```py
fn MyFunction[Requirements: Module](argument: Requirements):
    let x = Node[dtype](Tensor[dtype](1, 2, 3))
    _ = argument(x)

var h = nn.Linear(28*28, 256)
var a = MyFunction(h)
```

StijnWoestenborghs:
This whole thing is just not intuitive ...
```
fn call_one[T: Module](layer1: T, input: Int):
    print(layer1(input))

fn call_two[T: Module](layer1: T, layer2: T, input: Int):
    print(layer1(input))
    print(layer2(input))

# struct ALL[T: Module]:
#     var layer1: T
#     var layer2: T
    
#     fn __init__(inout self, inout layer1: T, inout layer2: T):
#         self.layer1 = layer1
#         self.layer2 = layer2


fn main():

    call_one(A(), 1)
    call_one(B(), 1)

    call_two(A(), B(), 1)

    # var all = ALL(A(), B())
    # print(all(1))
```
like why does call_one work and call_2 doesn't

andresnowak:
> This whole thing is just not intuitive ...
> 
> ```
> fn call_one[T: Module](layer1: T, input: Int):
>     print(layer1(input))
> 
> fn call_two[T: Module](layer1: T, layer2: T, input: Int):
>     print(layer1(input))
>     print(layer2(input))
> 
> # struct ALL[T: Module]:
> #     var layer1: T
> #     var layer2: T
>     
> #     fn __init__(inout self, inout layer1: T, inout layer2: T):
> #         self.layer1 = layer1
> #         self.layer2 = layer2
> 
> 
> fn main():
> 
>     call_one(A(), 1)
>     call_one(B(), 1)
> 
>     call_two(A(), B(), 1)
> 
>     # var all = ALL(A(), B())
>     # print(all(1))
> ```
> 
> like why does call_one work and call_2 doesn't

For me the call_two does work

StijnWoestenborghs:
Yeah i found you had to specify T2 as well. 
Anyway doing this variadic is something else. I might abandon the idea of doing the nn.Sequential

andresnowak:
yeah, it seems it is not possible for now, I thought of using like an append to a dynamic vector, but it isn't possible to use trait to define the value of the dynamic vector, so that also doesn't work.

StijnWoestenborghs:
Hmm good catch, but their are multiple things wrong here that need some investigating. 
Even with a softmax layer 8 the output tensor values are still all `inf`. (I secretly hope that that is the reason everything is that slow ... ). Something is going wrong with the in_channels of the maxpooling layer as well. I might have misinterpreted how that is handled. I'll be looking at it later. But feel free to push to it if you find something.

andresnowak:
okay, yeah I am also finding that in the softmax, the max values are 0 so they don't do any shifting, so maybei have done something wrong in the max max or the reduce function.

StijnWoestenborghs:
Probably a good way to tackle it is to find in what layer/op the issue(s) is and try to recreate it with a unit-test or something. A good chance that the speed is caused by `inf` values actually. 
(Extra: Probably the datasets.mojo deserves some refactor as well)

andresnowak:
- We first needed to use a RESHAPE (4, 32*28*28)before the linear layer (l7), because of how the dot operation is implemented (we need the tensors to be 2D)
- Second: The result values from the dot operation seem to big (we have values of of 4 million), (but maybe this is normal)
- Third: It seems there is something wrong in the LOG backwards operator, where the t tensor is not the same as the n tensor from the forward operation (t tensor has all its values set to zero, and from here we get a division by zero in the backward pass) **This seems is where the error is happening, or at least one of them**

StijnWoestenborghs:
- The reshape is needed ineed, essentially pytorch is also doing this using `x = x.view(x.size(0), -1)` So fixed (But API should change in the future)
- Result values are remain in the right order of magnitude now. The exploding values were caused by wrong initialisation of the weight and biases in the Linear and Conv2d layers. 
- I like checking the shape in elwise_op. So I removed all the double shape checks that were present in the operators. 
- On the CrossEntropyLoss. I changed the orignial implementation (following a tinygrad example). This is also included the LogSoftmax. Looks like this removes the need for this epsilon stability parameter. I didn't include a test yet, so if you have one comparing it to PyTorch cross entropy. Would be great to add it.
- Loss still remains quite stable though across epochs. I wonder if the parameters of the graph are actually updated...

andresnowak:
Yeah, let me commit the test. But I also tested using a different weight initializer (HE_uniform, with 1e-5 values), and the error would start around 2 and in the torch example would go down, but the mojo example instead goes up to 10 and stays there.

StijnWoestenborghs:
Training!
Cross entropy is fixed and produces similar output as torch.nn.CrossEntropy. You were right on the dim(0). Changed the loss function to use our LogSoftmax from activations.
The issue was that we forgot one-hot encoding in the MNIST example. You had it in your test

andresnowak (approved):

andresnowak:
Sorry I didn't remember to add to the readme that mnist ist completed

PR #17, opened 2024-01-26
Fix library to be able to run in mojo v0.7.0

andresnowak:
# Small fix to be able to run the library in mojo v0.7.0

## What was fixed:
Now if you want a function that can accept a VariadicListMem or the new Reference you need the syntax (for now):
```py
fn foos[lifetime: ImmLifetime](args: VariadicListMem[String, __mlir_attr.`false`, lifetime]):
    pass

fn bar(borrowed *args: String):
    foos(args)

fn main():
    var a = "hello"
    bar(a)
```
This are the attributes:
- type (AnyType): Type of the underlying data.
- ​is_mutable (i1): Whether the referenced data may be mutated through this.
- lifetime (lifetime<is_mutable>): The lifetime of the reference.

So the first one tells us the type, the second one if its mutable (but using mlir instead of mojo types) and the third one I think refers to the lifetime of the value (also saying if its mutable or not) like in rust (`a). So for now it is very verbose

And also remove the use of __get_addres_as_lvalue because it is not necessary anymore.

### DIFF
--- a/dainemo/autograd/graph.mojo
+++ b/dainemo/autograd/graph.mojo
@@ -21,10 +21,10 @@ struct Graph[dtype: DType = DType.float32, tracking: Bool = True](Stringable):
         self.keys = DynamicVector[String]()
         self.graph = DynamicVector[Node[dtype]]()
 
-    fn add_edge(
+    fn add_edge[lifetime_op: ImmLifetime](
         inout self,
         inout result_node: Node[dtype],
-        operands: VariadicListMem[Node[dtype]],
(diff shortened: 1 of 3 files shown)


StijnWoestenborghs:
Nice, will merge this into the MNIST as well.
PS: Have you seen the performance on the regression example? Performance has increase **A LOT** for me.

StijnWoestenborghs (approved):

andresnowak:
Yeah I thought it was my imagination, but i also thought that it looked faster. Maybe accessing values is faster now and some memory copies were removed

PR #18, opened 2024-02-27
Static Graph refactor

StijnWoestenborghs:
**Goals of this refactor:**
Static graph compilation

Other considerations:
- Complex operations: If one wants to implement softmax, but we implement with a graph instead of a manual implementation. Maybe it could be possible to make it have a backward and forward call, and both of them create a graph (Sometimes there can be a simpler backward operation).
- Op attributes: Add ability to have operator attributes (like conv needs padding, stride, etc)
- Broadcast shapes: maybe it would be better to do the broadcasting in the graph, Like what pytorch does where it creates another "tensor" (but it has a reference to the same memory)
- Symbol id names: Have ability to add a name to the result of an operation or of an input or param or output, to be able to access easily to the information of the tensor data.


andresnowak:
For the backward test I did them doing the call manually to the operators backward functions without using the graph, because to to do the test using the graph we would need to be able to give to the input gradients (or that's the only way I could think of) and I don't know if you would want that, so I just added tests for two operators for now to know your opinion.

StijnWoestenborghs:
I think there are two crucial functionalities left that are missing: 

- **Parameter initialization**: Something like `g.param(shape, init=Tensor(shape))` The problem with this is that we'll need to store these initialised Tensor somehow so that it passes the compiler. But maybe we can achieve this with a similar hack as attributes and have them stored in a Dict just like in constants. I'll have a look at this. Having these functionality will also allow for loading already trained models and only do inference. 

- **Loss**: We need a way to better split the loss function from the graph output. What do you think of this API? 
```
    var g = Graph()

    var x = g.input(TensorShape(batch_size, n_inputs))
    var y_true = g.input(TensorShape(batch_size, n_outputs))

    var W = g.param(TensorShape(n_inputs, n_outputs))
    var b = g.param(TensorShape(n_outputs))
    var res = g.op(OP.DOT, x, W)

    var y_pred = g.op(OP.ADD, res, b)
    _ = g.out(y_pred)
    
    var loss = mse(g, y_true, y_pred)
    _ = g.loss(loss)
```

And the the `Model` would two separate methods for a forward pass:
- One that goes all the way to loss, for a forward pass during training `loss = model.forward(input, labels)`
- One that only does a forward pass up to the output. `pred = model.inference(input)`

andresnowak:
I like the idea of the api to divide the loss using a special value (instead of out we use loss). But for the param initialization, wouldn't it be better to maybe use something like the attributes, like a symbol can also have attributes, and in that we pass the type of initialization that we want (Like kaiming) and the attributes that it would need? (like the fan in).

StijnWoestenborghs:
So this adds the functionality for parameter initialization. The big changes are that the `Constants` struct have been removed from the graph. Symbol has its requires_grad renamed to trainable. 
- So if you now want to add a constant/scalar to the graph you can do so by `graph.scalar` or `graph.constant`. And it gets stored as un-trainable parameter of the graph.
- And if you want to add a parameter with a certain init value you can do it with `graph.param(shape, init=Param(...))`. And this gets stored as a trainable parameter in the graph. 

The thing is that it is pretty hard to create a Param(...) because of comp time. As the whole random lib and Tensor are not allowed. The only way to define them (as far as I know) is using DynamicVector. Meaning that if we want our layer API's to create these random Param we'll need our own random generator. (which we actually have already in uuid)

StijnWoestenborghs:
Actually, we can probably support both ways. Param( dynamic_vector ) for inference only when you want to load your already trained model or Param( fan_in ) for the convenience when training.

andresnowak:
If both ways can be done then awesome, I just still for the training part maybe I didn't understood correctly, but would it be possible to define also other type of initialization function?, that's why i was thinking about using attributes to be able to do that, or would it also be possible or do you think it isn't necessary.

StijnWoestenborghs:
I think it should be possible here: https://github.com/StijnWoestenborghs/dainemo/blob/static-graph/dainemo/nn/model.mojo#L265-L273
If we add attributes to the `ParamDict` we will be able to reach them when allocating memory on initialization of the Model. So something like, Attribute("type", "kaiming"), Attribute("fan_in", val) could define all the inputs required. 

So in this case it'll look something like:
`g.param(TensorShape(n_inputs, n_outputs), init=Param(attributes=Attribute(...)))`
or
`g.param(TensorShape(n_inputs, n_outputs), init=Param( <dynamic_vector> ))`

andresnowak:
Okay so the param value can also save attributes and then in the model initialization we could use the information to do the initialization, I like the idea.

StijnWoestenborghs:
Had some fights with the compiler here but eventually got it to work reliably. I ended up with these API's
`g.param(shape)` -> defaults to initialising the parameter to 0
`g.param(shape, init="kaiming_normal")` -> using a specified initializer from `nn.initializers.mojo`
`g.param(shape, init=Param(<dynamic_vector>))` -> specifically setting the parameters

It is still using Attribute underneath but it is hidden from the API. And I tried to do this using AttributeVector but that just doesn't work consistently somehow. Also, for the initializers: apparently you don't need to supply the fan_in because you can derive it from the shape. So that is why I could still do it with only 1 attribute (Based myself on pytorch for this: https://github.com/pytorch/pytorch/blob/main/torch/nn/init.py)

andresnowak:
Awesome I like it. And huh I didn't know you can derive the fan_in I always thought that sometimes it was necessary to specify it.

StijnWoestenborghs:
So you are able to go for a full forward pass and for inferencing the model only now. And while I was there I added the ability to define multiple outputs. Some notes:
- Added some [warnings](https://github.com/StijnWoestenborghs/dainemo/blob/static-graph/dainemo/nn/model.mojo#L73-L77) for when your are only training and not defining the g.output and only inferencing and not defining the g.loss. Not sure about this but it's hard to enforce properly and not lose performance. 
-  Discovered a [pitfall](https://github.com/StijnWoestenborghs/dainemo/blob/static-graph/dainemo/nn/model.mojo#L82-L91) when defining the inputs. Again hard to enforce but maybe not too bad I guess. 

We can probably start wrapping it up now. Some layer API's, the MNIST example, cleaning up tests... And that is it 🔥

andresnowak:
It seems the thing that is slow right now is the conv operation (its forward and backward) and specifically the second conv operation is the one that takes the majority of the time, but I don't understand why in the present library example the conv operation isn't slow and why here it is very slow.

StijnWoestenborghs:
That is some strange behaviour indeed. I'm getting about a 20% performance gain when you are unpacking all the lists at comp time (w.r.t. having the static list at runtime). So maybe that is the way around this (not the most beautiful one though) 
```
        alias inputs_strides = calculate_strides(input_shape)
        
        alias inputs_strides_0 = inputs_strides[0]
        alias inputs_strides_1 = inputs_strides[1]
        alias inputs_strides_2 = inputs_strides[2]
        alias kernel_strides_0 = kernel_strides[0]
        ....
```

With alias unpacking around `11.5 sec` per epoch, strides at runtime is about `14 sec` per epoch.

StijnWoestenborghs:
Apart from this, it looks like a bug that is left is that it is not training using the CrossEntropyLoss (and only a little using MSELoss). So I would propose to fix this, merge, and open a new PR for the performance of the CONV2D op with proper benchmarking. So taking vectorizing and parallellization here could be done in parallel with other stuff.

Let's get this massive refactor merged first 😅 and go back to smaller feature updates

StijnWoestenborghs:
The following is the test that initializes similar weights and compares to a similar pytorch implementation. 
```
loss_mojo:  2.3025851249694824  loss_torch:  2.3025851249694824
loss_mojo:  1.3889272212982178  loss_torch:  1.390902042388916
loss_mojo:  1.3893520832061768  loss_torch:  1.3859763145446777
...
```
So, given the fact that the first loss is exactly the same. That leaves that he error could be either in the backward or the optimizer. (comparing the gradients could give us if it is in backward or optimizer)  I'll try to pinpoint exactly where it is deviating later.

andresnowak:
Finally fixed the problem with mnist not training, again sorry an error on my part. Unbroadcast_add was using vectorization and that was giving it errors (I have to find a way to do vectorization), and that was the problem (this would have been found before if I would have added a test, but now we have one). I discovered the error because I was thinking what was the difference with mseloss and cross entropy and a difference was that we were using broadcasted operations and from there came the idea.

StijnWoestenborghs:
Nice, I reorded/renamed the old unbroadcast_add there a little to make it a little more clear. Aside from the vectorization, I think the implementation is correct. It is training, yet still I'm puzzled why it is not converging in a similar way as torch. As far as I now we are not missing other concepts here. So I guess I will dive into comparing the gradients one on one.

andresnowak (approved):

PR #20, opened 2024-03-02
Idea for operation attributes

andresnowak:
Idea for how to how attributes in the static graph.

**Problems**
- The creating an attributeVector with attribute value or more, the program starts to have weird behaviors, like it seems memory pointers of objects and variables get moved around at compile time and we start to get weird behaviors that crash the program or just cause errors at runtime. **Problems with string** at compile time seem to have been fixed but there are still a lot of weird behaviors with lifetimes at comp time.

### DIFF
--- a/dainemo/autograd/graph.mojo
+++ b/dainemo/autograd/graph.mojo
@@ -2,7 +2,7 @@ from python.python import Python
 from tensor import TensorShape
 from collections.optional import Optional
 
-from .node import Node
+from .node import Node, AttributeVector, Attribute
 from .symbol import Symbol
 from .ops import OP, static_result_shape
 from .constant import Constant, ConstantDict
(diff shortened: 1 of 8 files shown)


StijnWoestenborghs:
Okay, I was doing the SUM operation myself and bumped into this requirement. (But saw very late you were working on this already with these attributes). 
- Wanted to include an axis field in the ops.mojo OP struct --> but this is not general for other operators like padding stride in conv.
- I feel like the attributes should be part of the OP struct --> but impossible because we can't really do something dynamic that is register passable and still generic for multiple different operators.

So as far as I can think of this idea of attaching the attributes to the node is (the only) way to go.
But still I feel like this will require refactoring in the future.

StijnWoestenborghs:
I'm probably going to try to merge this into the static_graph branch. wdyt?

andresnowak:
I think I understand when you say you think the attributes should be part of the struct, but I feel that having the attribute list in the node could give like more freedom (because attributes can be a lot of stuff) in what we could do but maybe I'm wrong.
And yeah I also think this is the only way we can go for now, but do you think you can fix the issue is happening, because right now having the attribute vector causes a lot of bugs

StijnWoestenborghs:
Absolutely, I will have a better look at it tomorrow. And actually I agree on having it as part of Node, more flexibility this early now is most likely a good and future proof thing.

PR #22, opened 2024-03-19
Update UUID implementation

StijnWoestenborghs:
- Updates MersenneTwister random generator to a struct that keeps state
- Strips uuid from using bytes and moves bytes from 
- Renames UUID type and UUIDGenerator

### DIFF
--- a/basalt/autograd/attributes.mojo
+++ b/basalt/autograd/attributes.mojo
@@ -2,7 +2,7 @@ from collections.optional import Optional
 from math import min
 from tensor import TensorShape
 
-from basalt.utils.uuid import bytes
+from basalt.utils.bytes import bytes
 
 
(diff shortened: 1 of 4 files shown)


Benny-Nottonson (commented):
lgtm

Benny-Nottonson (approved):

StijnWoestenborghs:
Only very minor changes compared to your rework. 
Some for readability, and I brought back `alias W: Int = 32` as there is a 64 bit variant of this. 
Did you do something to the unrolls?

Benny-Nottonson:
They dont really make a performance difference in my testing, not worth

StijnWoestenborghs:
Yeah okay, doesn't matter. I think they are in now

PR #23, opened 2024-03-19
Custom Tensor and TensorShape

StijnWoestenborghs:
- Define custom Tensor & TensorShape (holding on to std lib API's) 
~~- Biggest difference is that the new `Tensor` holds its shape as parameter. Simplifying runtime behaviour.~~
- Refactor basalt to use the custom Tensor and TensorShape


StijnWoestenborghs:
@andresnowak Can you take a look at this and verify if I'm not missing something in here, before implementing.
I think if we just change the interface of how the tensor are created in nn.Model during `allocate_tensor_memory` and how the tensors are stored in the Collection in the model.parameters. This can potentially clean up the signatures of `forward_op[t_shape, ...](res, t ...)`, `backward_op[ug_shape, t_shape...](ug, t ...)`  and all the ops as well. We can probably remove all the shapes as parameter.

Strides can be aliased from TensorShape and is returning the same result as we had before. Example:
```
var strides = calculate_strides(t_shape)
```
could be done
```
alias strides = t.shape.strides()
```
So even though shape is not passes as parameter in the forward pass, the tensor variable can hold the shape from which the strides can be aliased. I think it is pretty neat.

andresnowak:
From what I understand yeah the idea should work correctly. Line allocate tensor memory and in grad memory instead of using a for loop we would need to use an unroll loop to be able to get the compile time shapes and give it to the created it tenors as parameters (but the way we save it in collections can be the same from what I understand, the compile time value is saved in the struct and you can always access it in any way as compile time from what I understand). And then in the forward loop and backward It shouldn't be necessary to unroll the loop to access the shapes as comp values, but  if one wants to access the operators and the other information of the graph as comp time this is still needed (so forward and backward still need unroll). But yeah the only thing from what i understand would change is using unroll in allocate tensor memory and then one can remove in the forward and backward ops the need to pass the tensors as parameters to the function (unless there is some comp time bug).

StijnWoestenborghs:
Oh yeah that is actually a big thing I didn't think of. If this unroll becomes a problem at some point, we can totally avoid it by saving the comp time information (the attributes) in the collection somehow. And only unroll once at init, tot get the it out of the comp time graph.

andresnowak:
yeah, like for now we should keep the unrolling, and then see like at what point there are problems but at the end you need unrolling at least for now to get compile time information (so doing the kernel fusion probably and also to be able to give the compile time shapes to the tensors).

andresnowak:
I'm still not sure about using register passable in the tensor, seeing the complexity it adds to be able to manage the memory correctly and also that you have to manually free the memory of the tensor (Because register passable("trivial") value doesn't have a __del__ operator so you have to manage the memory manually and that they also have their own copy methods. I didn't really now this until I saw a post of michael kowalski.). And also testing the pocs of matmul and testing with different algorithms I don't see a difference in speed.

StijnWoestenborghs:
Still fixing up the last elements but some changes here:
- As you indicated the Tensor is not register_passable now. As far as i know, their shouldn't be any problem or performance difference, since as your tensor data is register passable. (Maybe even on the contrary, you might want to be careful not to fill the registers with data other then the actual tensor_data for optimized ops. But totally not an expert here...)  
- TensorShape is (just like the std tensor) not a parameter. I couldn't find a way to create a collection that could store Tensors in that way. Or couldn't find a way to work around this. This means model does require the `unroll[fn, count]()`. Closer to what we had before, but I did a test en for large number of nodes, compilation time just explodes. But we'll still have the TensorShape propagated as parameter to the operators, so I don't think it has an impact on performance.  
- Based on that post you mentioned I have found a way to have references in the collection! And I also integrated the stringdict inside the collection so you can reference the Tensor by Symbol. No `__get_as_lvalue(...)` anymore, no copies, and the string_dict is hidden. This cleans up the repo quite a bit.

basalt/autograd/attributes.mojo
```diff
@@ -37,13 +36,12 @@ struct AttributeVector(Sized, Stringable, CollectionElement):
     fn __str__(self) -> String:
         var s: String = "["
         for i in range(self._size):
-            s += str(self._attrs[i])
+            # s += str(self._attrs[i])
```
andresnowak, line 39: I think it is possible to still have these just pointing it out.

andresnowak (commented):
Everything seems okay. It's amazing how much readable the code now is, you really reduced some complexity. Only thing if you would want to check out is a comment i made, where it seems you commented a line that wasn't necessary to comment out I think, that's the only thing I could find really find. If you say its okay or if you change it i will approve it.

StijnWoestenborghs:
Voila, as I mentioned. There are some copies being done left on the Tensor. But they are more related to the dataloader and input/output handling of the graph. So would keep that for a next PR.

basalt/autograd/attributes.mojo
StijnWoestenborghs: Absolutely!

andresnowak (approved):

PR #25, opened 2024-03-24
Add profiler

Benny-Nottonson:
This pull request adds a profiler to the codebase. The profiler allows for performance analysis and generates a flame graph visualization.

### DIFF
--- a/profile.sh
+++ b/profile.sh
@@ -1,6 +1,39 @@
-# Usage . profile.sh <mojo_file>
+#!/bin/bash
 
 function profile() {
+    if [ ! -d ~/FlameGraph ]; then
+        InstallFlameGraph
+    fi
+
+    case "$OSTYPE" in
+        darwin*)
+            profileMac "$1"
+            ;;
+        linux-gnu*|msys)
+            profileLinux "$1"
+            ;;
+        *)
+            echo "Error: Unknown OS"
+            exit 1
+            ;;
+    esac
+}
+
+function profileLinux() {
+    local mojo_file=$1
+    LinuxInstallDependencies
+    LinuxPermissions
+    runProfile "$mojo_file"
+}
+
+function profileMac() {
+    local mojo_file=$1
+    MacInstallDependencies
+    MacPermissions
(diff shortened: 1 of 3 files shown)


StijnWoestenborghs:
Are you on WSL for this? As I'm on Linux and if we are going to add a profiler to the main branch we should make it work for all platforms probably. The changes I had to make were:
```
... "$temp_dir/run.exe" --> "$temp_dir/run"
~/perf -> sudo perf 
explorer.exe --> google-chrome $flamegraph_output
```

I'll have a look later, for the script to recoginze the platform. 
@andresnowak  Are you on Mac perhaps?

Benny-Nottonson:
Good point about being platform independent, im on WSL2 Ubuntu, I also have a Mac I can test on if needed, but I agree, we should unify that for all platforms. The ~/perf is a result of a workaround needed to install perf on linux, and could probably be put into a seperate step, ill look into detecting platform.

Benny-Nottonson:
@StijnWoestenborghs Check now :)

profile.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    echo "Building $mojo_file..."
+    mojo build -O0 -I . "$mojo_file"
+
+    echo "Stripping debug symbols..."
+    mv "$mojo_name" "$temp_dir/run.exe"
+    llvm-strip --strip-debug "$temp_dir/run.exe"
+
+    echo "Running perf record..."
+    perf record -F 99 -a -g -o "$perf_output" -- "$temp_dir/run.exe"
```
StijnWoestenborghs, line 55: For some reason i still need a sudo here to get it to work. 
And does it work for you without the .exe (=windows specific)

profile.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    echo "Generating flamegraph..."
+    perf script -i "$perf_output" | ~/FlameGraph/stackcollapse-perf.pl | ~/FlameGraph/flamegraph.pl > "$flamegraph_output"
```
StijnWoestenborghs, line 58: same here, needs sudo

PR #28, opened 2024-03-25
Add tests using torch to compare

andresnowak:
- [x] Add forward and backward ops tests
   - [x] Fix error in div broadcast backward
   - [x] Fix error in vectorized reduced ops
   - [x] Fix error where we weren't calling the backward mean operator passing it it's attributes vector
   - [x] Fixed error in the to_tensor function (The one used to convert numpy to basalt tensor) that was making it really slow
- [x] Add forward and backward mlops tests

P.S: There can be sometimes a lot of difference in decimal values, but probably the problem is with decimals not being deterministic when using parallelization.


StijnWoestenborghs:
All good to me!

StijnWoestenborghs (approved):

PR #29, opened 2024-03-26
Replace DynamicVector in ops

StijnWoestenborghs:
Exploring the possibility of avoiding the call to the general `(un)broadcasting` functions, and specifically focus on tensors of rank 1 or 2 (as in the Linear layer). As shapes are statically known, broadcasting & unbroadcasting could end up in another entrypoint without any overhead.

This would have the possiblity of reducing the overhead on `get_real_index` in `broadcast_shapes` and `accumulate_grad` as seen in the `sin_estimate.mojo` benchmark, which contains multiple Linear layers

### DIFF
--- a/basalt/autograd/graph.mojo
+++ b/basalt/autograd/graph.mojo
@@ -112,7 +112,7 @@ struct Graph:
             )
 
         self.nodes.push_back(
-            Node(op, res, operand_1, operand_2.take(), operand_3.take(), attributes)
+            Node(op, res, operand_1, operand_2, operand_3, attributes)
         )
         self.symbol_count += 1
         return res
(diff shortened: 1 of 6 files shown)


StijnWoestenborghs:
**Update**: I hacked inside the broadcasting function and implemented something optimised for the case where subtracting the bias (rank 1 of dim n_outputs) from the result of the dot ( batch x n_outputs ) is triggering the broadcasting function. 
This significantly reduced the amount of calls to `get_real_index` but didn't have any noticeable effect on performance. 
Abandoning the main idea here.

The changes left are:
- Fixing a bug on graph render
- Speeding up slightly by throwing any dynamic vector and replacing with static ones in the operators

PR #38, opened 2024-04-01
Added the equalityComparable trait to symbol

Benny-Nottonson:


Benny-Nottonson:
Gonna merge this since we still need to review cleaning, is inline with what is mentioned about trying to explicitly conform to traits.

PR #40, opened 2024-04-02
Add a simple perf metrics module for model

andresnowak:
This pull request adds a simple performance metrics to model. To be able to check the speed of each operator in the forward and backward pass, how much percentage of the total time each operator is taking and also print the shape if the user wants for more information in debugging.

The only problem with this implementation is that one needs to modify the compile time DEBUG variable manually, using the command mojo -D DEBUG=1 -I . "file", causes an error at runtime that seems to be related with linking (so it seems the error is because of using -D and -I together).


StijnWoestenborghs:
This is totally awesome. Let's work it out a little better, things I can think of:
- Move to utils so it doesn't clutter nn.model too much. 
- Optional member of model
- Increase readability of results
- Is there a way to print this automatically .. ? Maybe a new method: `model.perf()` to print the results
- It seem so be breaking the housing benchmark
I'll have a look as well

andresnowak:
I'm seeing the the error in housing example very strange. Also if you didn't know but I think this is an error of the language not us, but in mac the examples get errors, like for housing we get a nan error, sometimes since the first forward and sometimes after a backward, in the mnist example an error happens where the error doesn't go down from 2.3, but the tests run correctly, but it is really strange.

the other thing what did you want to say with print automatically. just having a method in model that does the print instead of calling the function from the per_metrics variable.

StijnWoestenborghs:
Interesting, I meant a way that you don't have to print the metrics manually when setting the environment variable. Like that at the end of you program/loop the statistics get printed. I didn't saw an immediate way of doing that so that is why an alternative could be a `model.perf()` method. Or maybe something else... still not entirely sure about this.

andresnowak:
I think it is better to do it manually, like it is more the option to the user what it wants to do, like if they want to print also shapes or if they want the time in ms, s, ns, etc. But I think the `__del__` method accepts prints

andresnowak:
So guess what the error was, the error was the dataloader, the last batch was grabbing values that were outside the allocated values for the data and labels. So it seems depending on luck, the memory values garbage values were 0 instead of a random number, and probably this was the error in mac. It isn't the first time i have seen that in linux there is luck were the garbage values are 0 instead of a random value, I don't know why that happens

StijnWoestenborghs:
Oooh, nice catch! Actually this is leading to better (/solving the) convergence on both the housing and mnist benchmarks. In housing, I noticed we went from 22 to 29 loss. And now were back to 22 again (same as torch). I have no idea why this didn't cause any further issues before, clearly a bug.

andresnowak:
So optional seems that can't be used, because when the start_forward_pass and end_forward_pass you are modifying the struct, and it seems that optional doesn't like for the value that it contains to be modified, the other thing would be to always  __take__ the value and then give it back, but i think that makes a copy. And for the formatting i don't know what types of things you wanted, like i did some changes but i don't know if you wanted other stuff, like only print 3 decimals, but the thing is right now the language doesn't have that, i could try and do it but i don't know how effective it would be.

StijnWoestenborghs:
@andresnowak Spend some time on a pretty string format

StijnWoestenborghs:
Some nice insights already:
In the MNIST, the bw SUB seem to be taking the majority of the time. Possibly due to the accumulate grad & de-broadcasting

andresnowak:
What i can i say, better than what i could have done, i always have a difficult time also thinking about the fit string thing, even though it is something not super complicated, I do really like it.

StijnWoestenborghs:
Hmm, I just see that `truncate_decimals` needs some improvement as you might end up with a `0.000` now.

andresnowak:
oh but one can change how many decimals they want no?, maybe another option that can be passed to the formatter?, how many decimals one wants?

StijnWoestenborghs:
Yeah I just increased it to 4 for time. It mitigates the problem, but I think it's not worth spending more time on these details

andresnowak:
Well that's true yeah it is already very readable and it serves its purpose of giving performance information

StijnWoestenborghs:
Cool, good for me

PR #43, opened 2024-04-05
Added FMA Operator

Benny-Nottonson:
Added the FMA (Fused Multiply Add) operater to the OP struct, added associated tests.

### DIFF
--- a/basalt/autograd/ops/basics.mojo
+++ b/basalt/autograd/ops/basics.mojo
@@ -711,3 +711,52 @@ struct RESHAPE:
         memcpy(res_grad.data(), ug.data(), ug_shape.num_elements())
 
         return res_grad ^
+
+struct FMA:
+    @staticmethod
+    fn result_shape(t1_shape: TensorShape, t2_shape: TensorShape, t3_shape: TensorShape) -> TensorShape:
+        return broadcast_shapes(t1_shape, t2_shape, t3_shape)
+
+    @staticmethod
+    fn forward[
+        t1_shape: TensorShape,
+        t2_shape: TensorShape,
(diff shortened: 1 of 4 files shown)


basalt/autograd/ops/basics.mojo
```diff
@@ -711,3 +711,52 @@ struct RESHAPE:
...
+        @parameter
+        if tensor_id == 0:
+            var res_grad = Tensor[dtype](ug_shape)
+            elwise_op[ug_shape, t2_shape, mul](res_grad, ug, t2)
+            return res_grad ^
+        elif tensor_id == 1:
+            var res_grad = Tensor[dtype](ug_shape)
+            elwise_op[ug_shape, t1_shape, mul](res_grad, ug, t1)
+            return res_grad ^
+        else:
```
soraros, line 761: Should we put a `constrained[False, ...]()` here?

Benny-Nottonson: Maybe, but I do think we need a wider discussion about how we want to manage Error handling, especially when we get to more constrained operators like Cast

soraros: For `cast`, I thin constrain can be imposed like `stdlib.SIMD`. `constrained[False]()` is really poor man's unreachable. I'm good with not addressing error handling here.

andresnowak:
I think everything looks good. Just the only thing missing, if you can also add a test in test_ops_torch, that's the only thing missing I would say.

Benny-Nottonson (commented):
looks good to me

andresnowak (approved):

PR #44, opened 2024-04-05
[BLOCKED] Add CLIP operation

Benny-Nottonson:
This pull request adds a new CLIP operation to the codebase. The CLIP operation allows for clipping the values of a tensor within a specified range. This can be useful for tasks such as gradient clipping in neural networks.

### DIFF
--- a/basalt/autograd/ops/basics.mojo
+++ b/basalt/autograd/ops/basics.mojo
@@ -1,7 +1,6 @@
 from math import add, sub, mul, div, log, exp
 from algorithm import vectorize
 from memory import memcpy
-from math.limit import max_finite
 
 from basalt import Tensor, TensorShape
 from basalt.nn.tensor import max_rank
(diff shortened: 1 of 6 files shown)


soraros:
Can we implement it as `min(max(x, ub), lb)`?

Benny-Nottonson:
> Can we implement it as `min(max(x, ub), lb)`?

Wouldnt that be the same as `x.max(ub).min(lb)` ?

soraros:
Yea, my point is, do we need a new Op?

Benny-Nottonson:
> Yea, my point is, do we need a new Op?

It may be a little bit of syntactic sugar but I think its cleaner to give it as an option rather than forcing max and min seperately

soraros:
I'm definitely not against providing wrapper functions with clear semantics. I think we should have a concept of "compositional op", so we reuse the impl and AD.

Benny-Nottonson:
> I'm definitely not against providing wrapper functions with clear semantics. I think we should have a concept of "compositional op", so we reuse the impl and AD.

I don't see the inherent benefit in implementing this specific function as a composition of Min and Max since it is fairly simple, but im open to everyone elses thoughts on this.

andresnowak:
Sorry for my ignorance but when saying compositional op, what we want to say is that CLIP should be ```x.max(ub).min(lb)```, instead of making it a primitive operator?.

andresnowak:
Okay. Then for me the only thing missing is that I would say this operator should be in mlops instead of basics, And to change the test from test_ops to test_mlops, and to also have a test_mlops_torch.

PR #48, opened 2024-04-08
Attributes accepting Scalars

StijnWoestenborghs:
- Reworking attributes to accepts scalar values of a given dtype
- Unblocking: CLIP: https://github.com/basalt-org/basalt/pull/44

NOTE: The current implementation of AttributeValue is a workaround.
Can't seem find a way so far to support attributes of Variant types that are able to pass through the information from comp time to the ops. Limited bytes support requires a custom bytes implementation.


StijnWoestenborghs:
Stopping in favor of:
https://github.com/basalt-org/basalt/pull/49

PR #49, opened 2024-04-09
Rework of Attribute

Benny-Nottonson:
Upgrades current implementation by making casting Scalars better, removing the need for two structs for a value, and improving code quality

### DIFF
--- a/basalt/autograd/attributes.mojo
+++ b/basalt/autograd/attributes.mojo
@@ -2,7 +2,7 @@ from collections import Optional
 
 from basalt import Tensor, TensorShape
 from basalt.nn.tensor import MAX_RANK
-from basalt.utils.bytes import Bytes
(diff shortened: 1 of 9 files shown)


StijnWoestenborghs:
Included a test here as I bumped into some issues. 
- When the Attribute is a variable (not alias). Getting a string from a the Bytes seems to fail
- I might be doing something wrong, but the attributes as scalar is segfaulting

Addition: Is it okay if we change `Type` to `dtype`  (for now! untill we are out of this) Just to keep consistency across the project. We'll change this after this discussion is cleared out

Benny-Nottonson:
thought I could sneak the Type bit past you, i'm good with using dtype for now 😉

Benny-Nottonson:
I'll have a look at the segfaulting if you haven't fixed in a bit

StijnWoestenborghs (approved):

PR #53, opened 2024-04-10
Add CLIP operation to autograd

Benny-Nottonson:
This pull request adds the CLIP operation to the autograd module. The CLIP operation performs element-wise clipping of a tensor's values between a specified minimum and maximum value.


Benny-Nottonson (commented):
Potentially needs aditional tests in test_mlops.mojo, but passing torch based tests.

StijnWoestenborghs:
Added some tests here. These are the issues left that I found. For sure the first two need to be fixed before merging.
(Note that I'm also casting an Int attribute to a scalar f64 now, as the previous implementation could only could go up to 255, uint8 max. But because of it, I'm getting that segfault on extracting the int at compile time)

```
from basalt.autograd.attributes import Attribute
from math.limit import max_finite, min_finite

fn main():
    alias dtype = DType.float32

    # Needed to continue
    alias a = Attribute("min", 0.3333)
    print(a.to_scalar[dtype]())  # 0?

    alias a2 = Attribute(name="axis", value=1)
    alias axis = a2.to_int()    # Segfault related to implicit casting types when shifting
    print(axis)

    # Not urgent
    alias b = Attribute("max", 1e16)
    print(b.to_scalar[dtype]())

    alias c = Attribute("min", max_finite[dtype]())
    print(c.to_scalar[dtype]())
```

StijnWoestenborghs:
Everything passing now except:
https://github.com/basalt-org/basalt/blob/clip_new/test/test_attributes.mojo#L103
Might be reasonable to continue though...

StijnWoestenborghs (approved):

PR #54, opened 2024-04-10
Added Squeeze Operator (Needs Tests)

Benny-Nottonson:
Added a basic implementation of Squeeze assuming Res is the correct shape, not sure how this is handled but I believe this is correct.


StijnWoestenborghs:
Closing in favor of: https://github.com/basalt-org/basalt/pull/56
Which has updated Attributes and Clip operators

PR #55, opened 2024-04-22
Implement q_sqrt algorithm

Benny-Nottonson:
This pull request implements the q_sqrt algorithm for calculating the square root of a number. The algorithm is used in the Conv2d and Linear functions to calculate the bound for weight initialization.

More info can be found here
https://en.wikipedia.org/wiki/Fast_inverse_square_root


StijnWoestenborghs (approved):
nice

PR #56, opened 2024-04-22
squeeze & unsqueeze operators

StijnWoestenborghs:
- Adding the squeeze operator
- Adding the unsqueeze operator
- Unittesting both

NOTE: Merging the old squeeze and unsqueeze branches here, as they had conflicting changes on Attributes

### DIFF
--- a/basalt/autograd/ops/mlops.mojo
+++ b/basalt/autograd/ops/mlops.mojo
@@ -209,6 +209,112 @@ struct CLIP:
         return res_grad ^
 
 
+struct SQUEEZE:
+    @staticmethod
+    fn result_shape(t1_shape: TensorShape, attributes: AttributeVector) -> TensorShape:
+        var dim = attributes["dims"]
+        var dims = attributes["dims"]
+
+        if not dim and not dims:
+            var new_rank = 0
+            for i in range(t1_shape.rank()):
+                if t1_shape[i] != 1:
+                    new_rank += 1
+            var new_shape = List[Int](capacity=new_rank)
+            for i in range(t1_shape.rank()):
+                if t1_shape[i] != 1:
+                    new_shape.append(t1_shape[i])
+            return TensorShape(new_shape)
+        elif dim:
+            var to_remove = dim.value().to_int()
+            var new_rank = t1_shape.rank() - 1
+            var new_shape = List[Int](capacity=new_rank)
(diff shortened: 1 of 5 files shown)


StijnWoestenborghs:
I think calculating the result shape of squeeze and unsqueeze can be simplified a lot, let me try

StijnWoestenborghs:
- Its maybe a little weird that specifying dim is stored as TensorShape in Attribute. 
- Supporting multiple dims in unsqueeze is a litlle more complicated as it would most likely assume a sorted list to work. Also interesting that [torch.unsqueeze](https://pytorch.org/docs/stable/generated/torch.unsqueeze.html#torch.unsqueeze) support only 1 value here and [jax.expand_dims](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.expand_dims.html) does support a tuple

andresnowak:
The only thing is that when one only passes one dimension (without using tensors shape or static_tuple), the way the value is saved in attributes is saved is different and the value can't be converted to tensorshape, at least not easily.

StijnWoestenborghs:
I think changing attributes is probably the right call here. If we do that, the API of attributes you just take in and return a Tuple/List or something. (also @Benny-Nottonson also wanted to get rid of `data_shape` in attributes completely.)

StijnWoestenborghs:
I think a hacky way now could be, changing attributes to:
```
    fn to_shape(self) -> TensorShape:
        if self.data_shape == StaticIntTuple[MAX_RANK]():
            return TensorShape(self.data[0].to_int())
        else:
            return TensorShape(rank=self.data[0].to_int(), shape=self.data_shape)
```
But I think its better to remove `data_shape` and `to_shape() -> TensorShape` completely from Attributes and have from list and `to_list` method that return a list of one if only one scalar/int was specified.

StijnWoestenborghs:
Alright for me this is OK actually. Having attributes store a list could be a separate PR imo.

Benny-Nottonson (approved):
I agree, looks good to me

andresnowak:
I just wanted to add a last test for unsqueeze with torch, but because unsqueeze behaves differently i don't know what other alternative to use, one is reshape but i don't know if to use another one.

andresnowak:
Okay now i think it can be merged

andresnowak (approved):

PR #57, opened 2024-04-24
CONCAT2 Operator

StijnWoestenborghs:
Concatenating 2 operand tensor along a specified dimension
- CONCAT2 forward op
- CONCAT2 backward op
- unittests for both

**remark 1**
We need to fix the number of operands in the core operator here as a requirement to statically know the shapes. The API could expose a more general `concat()` function taking in a list of symbols of arbitrary length. And that operator can use a series of core CONCAT2 ops, or implement more CONCATx 

**remark 2**
I was originally planning on implementing split here as well, but this operator is going to be more complex then originally though. So focussing on concat only. The graph doesn't have issues with specifying Nodes/OPs that have multiple outputs. But the flow in model.execute is going to be WAY more complicated if we want to support core operators with multiple outputs. My current thought on SPLIT is to implement SLICE as core operator and create `g.split()` as a combination of slices. That should be equivalent as slice fw/bw & accumulate_grad should take care of everything. Partially proposing this as well as (afaik) split and slice are the only ops with multiple output tensors in the graph. In the case that is true, it's probably not worth complicating model.execute that much for only these cases. Thoughts are more then welcome on this ...

### DIFF
--- a/basalt/autograd/graph.mojo
+++ b/basalt/autograd/graph.mojo
@@ -78,8 +78,11 @@ struct Graph:
         operand_3: Optional[Symbol] = None,
         attributes: AttributeVector = AttributeVector(),
     ) -> Symbol:
+        # TODO: Check if this can be *operands: Symbol
         var res: Symbol
+        var inputs: List[Symbol]
         if operand_3:
+            inputs = List[Symbol](operand_1, operand_2.value(), operand_3.value())
             res = Symbol(
                 self.symbol_count,
                 dtype,
@@ -93,6 +96,7 @@ struct Graph:
                 self.result_trainable(operand_1, operand_2.value(), operand_3.value()),
             )
         elif operand_2:
+            inputs = List[Symbol](operand_1, operand_2.value())
             res = Symbol(
                 self.symbol_count,
                 dtype,
@@ -102,6 +106,7 @@ struct Graph:
                 self.result_trainable(operand_1, operand_2.value()),
             )
         else:
+            inputs = List[Symbol](operand_1)
             res = Symbol(
                 self.symbol_count,
                 dtype,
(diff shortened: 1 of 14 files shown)


StijnWoestenborghs:
**remark 3**
There is also this idea of not initialising/copying new memory in ops like. squeeze, unsqueeze, concat, slice, split... But so far I haven't figured out how to realise that.

Benny-Nottonson:
first off, code looks good afaik, will run tests when i'm back home.

for remark 2: I think the slice method you describe is probably the best way to go, but that isn't necessarily my area of expertise 

for remark 3: this is huge and something i've been looking into, i think it's doable depending on how we manage graph computations by essentially only having one layer of the graph have data at once, you could theoretically use swap instead of memcpy, definitely needs more thinking but that's a seperate pr

changes lgtm

Benny-Nottonson:
also with regards to your first comment, we definitely do need a CONCATx so we don't end up in the same situation as Modular with external_call (see https://github.com/modularml/mojo/pull/2396/commits/4d4c2119799c42c29bd14a3ca8a72ce3e9feabd7)

StijnWoestenborghs:
**More thoughts on split:**
If we did want to implement split as a core op in `model.execute` I think this would be the affect:
We could implement a `SPLIT2` and assuming we only care about the case with 1 operand and 2 outputs.
An extra check on `if op == SPLIT2` in the forward & backward for `num_operands == 1`
https://github.com/basalt-org/basalt/blob/main/basalt/nn/model.mojo#L148-L150
https://github.com/basalt-org/basalt/blob/main/basalt/nn/model.mojo#L201-L208
In that case similar to CONCAT2, the API can expose a `g.split( ... )` for an arbitrary number of outputs.

StijnWoestenborghs:
Closing this one in favor of:
https://github.com/basalt-org/basalt/pull/58

PR #58, opened 2024-04-26
Dynamic ops

StijnWoestenborghs:
Quite a lot going on here:

**Big changes**
- Introduces **dynamic ops**: And the ability to mark an OP as dynamic. Which does still require the shape to be statically known for ops further in the tree but allows operators to have arbitrary number of inputs and number of outputs. (required for CONCAT & SPLIT)
-  ~~TENSORS and GRADS are now globals. (this is not a requirement for dynamic ops but makes things easier). If you would create a reference it would be still possible to not have them as members of model. Exploring this for convenience and for future graph modules. While not affecting runtime performance, I have the feeling compile time has increased because of this.~~   UPDATE: back to members of Model in model.parameters & model.parameters is being Referenced to other parts of the code 
- CONCAT: fw, bw as a dynamic op, with any number of operands. (and unittests)
- SPLIT: fw, bw as a dynamic op, with any number op outputs. (and unittests)
- Note that, for now, the API of those new ops is `g.split(...)`  and `g.concat(...)`. A different way of doing this would be `nn.split(g, ...)` and `nn.concat(g, ...)` like we did with the other layers. But leaving it like this as we probably want to keep it for the overall API discussion.

**Smaller updates**
- Some Graph cleanups
- Some broadcasting cleanups
- Some cleanups in testing ops
- model inputs are allowed to have gradients as well
- ability to specify upper gradients for `model.backward( * uppger_grads )`


basalt/__init__.mojo
```diff
@@ -1,6 +1,10 @@
 from .autograd import Graph, Symbol, OP
 from .nn import Tensor, TensorShape
+from basalt.utils.collection import Collection
 
 alias dtype = DType.float32
 alias nelts = 2 * simdwidthof[dtype]()
 alias seed = 42
+
+var TENSORS = Collection()
+var GRADS = Collection()
```
andresnowak, line 10: Why was it necessary to move the tensor grads from the model to global objects?

andresnowak:
Another thing is that it seems compile times have increased somewhat, like test_ops_torch takes a lot of time to compile, and seeing the code i don't know why that happened

StijnWoestenborghs:
Yeah so, first
- I don't think it is a necessity to have the tensor and grad memory as global. We can make them members of model again. And use references and lifetimes to access them in other parts of the code. 
- The immediate affect is that it clean up the API a little 
`optim.allocate_rms_and_momentum(model.parameters)` is gone and `optim.zero_grad(model.parameters)` and `optim.step(model.parameters)` doesn't need model.parameters anymore.
- It was convenient in dynamic operators as the forward and backward functions pass trough a list of symbols. So to get the actual tensor and grad, it could be done like this. But there might be a way to pass the model.parameters as reference to the kernel. 
- That being said, the idea came from submodules. If the loss function and the model would each be their own module (thus containing their own graph), API and nn.model could improve as each module would have their own forward and backward method. Still two problems with modules: 1) being that you need a way to specify what outputs of one module connect to the inputs of the other, at compile time. (we might need the ability to name symbols here)  2) without a global tensors and grads, this is an extra copy of the connecting tensors every fw and every bw pass (unless references?)

I also noticed the increase in compilation time on that test. That started to happen as soon as those global tensors and grads were in. So it might not be the right call. Its not a lot of effort to have them back as members of model btw, just struggling to find the optimal way here.

basalt/__init__.mojo
Benny-Nottonson: Another note on this, if we are using Collection for Tensors and Grads, how much performance overhead is caused by the Collection struct? I actually don't mind these being global but I do think we should make sure Collection is VERY optimized if we go this route.

andresnowak:
> Yeah so, first
> 
>     * I don't think it is a necessity to have the tensor and grad memory as global. We can make them members of model again. And use references and lifetimes to access them in other parts of the code.
> 
>     * The immediate affect is that it clean up the API a of the examples little
>       `optim.allocate_rms_and_momentum(model.parameters)` is gone and `optim.zero_grad(model.parameters)` and `optim.step(model.parameters)` doesn't need model.parameters anymore.
> 
>     * It was convenient in dynamic operators as the forward and backward functions pass trough a list of symbols. So to get the actual tensor and grad, it could be done like this. But there might be a way to pass the model.parameters as reference to the kernel.
> 
>     * That being said, the idea came from submodules. If the loss function and the model would each be their own module (thus containing their own graph), API and nn.model could improve as each module would have their own forward and backward method. Still two problems with modules: 1) being that you need a way to specify what outputs of one module connect to the inputs of the other, at compile time. (we might need the ability to name symbols here)  2) without a global tensors and grads, this is an extra copy of the connecting tensors every fw and every bw pass (unless references?)
> 
> 
> I also noticed the increase in compilation time on that test. That started to happen as soon as those global tensors and grads were in. So fairly sure it is the globals, so it might not be the right call. Its not a lot of effort to have them back as members of model btw, just struggling to find the optimal way here.

Okay I understand the points (_and sorry I didn't read the first comment (you already pointed out somethings there i don't know why i didn't saw it_), but yeah even thought it cleans the api a little bit and for the fourth point there is the thing with the forward and backward copies I think it is better to wait in this things for the language to get better (even if maybe at first if we want to implement them maybe they are sub optimal), because I think we can complicate ourselves when implementing things and also we have to do changes to the library that maybe we think aren't correct, that's my opinion at least.

basalt/autograd/graph.mojo
```diff
@@ -73,44 +73,27 @@ struct Graph:
...
-                ),
-                self.result_trainable(operand_1, operand_2.value()),
-            )
+        var res_shape: TensorShape
+        if len(operands) == 1:
+            res_shape = static_result_shape(op, operands[0].shape, attributes)
+        elif len(operands) == 2:
+            res_shape = static_result_shape(op, operands[0].shape, operands[1].shape, attributes)
+        elif len(operands) == 3:
+            res_shape = static_result_shape(op, operands[0].shape, operands[1].shape, operands[2].shape, attributes)
```
Benny-Nottonson, line 85: Not really relevant for this file, but is it possible to replace the static_result_shape call signature to take in *TensorShape instead of having 3 seperate overloads? Or even have one main overload that takes in *TensorShape and calls the appropriate version of the function?

Something like

```mojo
fn static_result_shape(op: Operator, *shapes: Symbol, attributes: AttributeVector) -> TensorShape:
    if len(shapes) == 1:
        return static_result_shape(op, shapes[0].shape, attributes)
    elif len(shapes) == 2:
        ...
    # etc..
    ```

File: basalt/autograd/node.mojo [hunk 3]
```diff
@@ -10,38 +10,35 @@ from .attributes import AttributeVector
 @value
 struct Node(CollectionElement, Stringable):
     var operator: OP
-    var output: Symbol
-    var input_1: Symbol
-    var input_2: Optional[Symbol]
-    var input_3: Optional[Symbol]
+    var inputs: List[Symbol]
+    var outputs: List[Symbol]
```
Benny-Nottonson, line 14: Have we considered a different type other than List since it has so much performance overhead?

basalt/autograd/ops/basics.mojo
```diff
@@ -718,7 +718,8 @@ struct FMA:
     fn result_shape(
         t1_shape: TensorShape, t2_shape: TensorShape, t3_shape: TensorShape
     ) -> TensorShape:
-        return broadcast_shapes(t1_shape, t2_shape, t3_shape)
+        # FMA assumes: t1_shape == t2_shape == t3_shape
+        return t1_shape
```
Benny-Nottonson, line 722: We should start commenting `# TODO: Error Handling` on places like these for when we eventually add some assertion logic.
```suggestion
        # FMA assumes: t1_shape == t2_shape == t3_shape
        # TODO: Error Handling
        return t1_shape
```

basalt/nn/layers/conv.mojo
```diff
@@ -8,7 +8,7 @@ from basalt.autograd.attributes import AttributeVector, Attribute
 
 @always_inline("nodebug")
 fn q_sqrt(value: Float32) -> Float32:
-    var y = bitcast[DType.float32](0x5f3759df - (bitcast[DType.uint32](value) >> 1))
+    var y = bitcast[DType.float32](0x5F3759DF - (bitcast[DType.uint32](value) >> 1))
     return y * (1.5 - 0.5 * value * y * y)
```
Benny-Nottonson, line 12: Since we are using this exact formula twice in two different files, we should probably extract it to a different file, can take care of when im home.

basalt/nn/model.mojo
```diff
@@ -121,20 +98,18 @@ struct Model[
         # TODO: known copies (reference?)
         var outputs = List[Tensor[dtype]]()
         for i in range(len(g.outputs)):
-            outputs.append(self.parameters.params[g.outputs[i]])
+            outputs.append(TENSORS[g.outputs[i]])
         return outputs ^
 
     fn execute[num_nodes: Int](inout self, t_input: VariadicListMem[Tensor[dtype]]):
```
Benny-Nottonson, line 104: Again, is there a way we can add an overload to take in a VariadicList of tensor collections instead of doing the logic here? Just thinking in the future if we have even more inputs.

Benny-Nottonson (commented):
Added some comments for related PR's / Small changes, but overall changes look good to me, tests / examples running on my Mac and WSL Ubuntu, slightly faster too :)

Benny-Nottonson:
I also think the overall API changes are moving in the right direction, this is definitely easier to use/read than before.

StijnWoestenborghs:
Yeah it is also slightly faster at runtime for me, however compile time ... (after this I also have some ideas about a real Frontend of the library, but other discussion).

But thus, to bo honest. I don't like TENSORS and GRADS being globals. It makes much more sense for them being part of Model right now (and probably Module in the future). I just think that right now, we will be punished in performance if we don't do it. My take
- CON: makes much more sense to be part of Model
- CON: compile time is bad.
- PRO: performance, (especially when looking at modules)
- PRO: easier to handle, readability and less lines of code

I'm actually more leaning to removing them again, and figure out a way for dynamic ops to work without them. As they are relying on it right now.

basalt/autograd/ops/basics.mojo
StijnWoestenborghs: Yeah, the API should handle this imo. To keep the kernels clean

andresnowak:
For me my opinion is also that I think that tensors and grads should be part of the model, And even if right now the performance can be better (like the example with tests right now it is probably faster because it is not allocating each time it runs a test) like for the module example in the future, I think many of those things is just better to wait for the language to mature (like references) in those part even if at first the performance wouldn't be the best at first i think many things will get better with updates. That would be my opinion for that part

StijnWoestenborghs:
Okay, then I'll be looking into how to do that, should be possible. I hope the dynamic ops are nice with me here. 
In any case. TENSORS and GRADS here were formerly called `model.parameters` which is a confusing name in itself.

andresnowak:
> Okay, then I'll be looking into how to do that, should be possible. I hope the dynamic ops play nice with me here. In any case. TENSORS and GRADS here were formerly called `model.parameters` which is a confusing name in itself, imo.

Really i thought it was okay because torch and other use that name, parameters are basically the tensors and grads no?

StijnWoestenborghs:
Oh right, but yes, I think our tensors at least is more, it also stores intermediate tensor memory between ops. Which not necessarily is a model parameter (that should be kept when saving and inference the model)

andresnowak:
oooh I think I understand, even though i don't think it is a problem having it named parameters i think it conveys somewhat the same idea, but i think I understand what you want to say

basalt/autograd/node.mojo
StijnWoestenborghs: Since this is comptime, the only alternative would be to have a `StaticTuple[Symble, MAX_...]()` with a fixed max length, I think. It also complicates knowing how many inputs and outputs you actually have in the node. We could try to see what the effect on performance is. But it would make to code worse i think.

andresnowak: I agree I feel the code would be worse and I also think List will get better in the future, i think those performance problems is better to wait, and even if in the future there is a little overhead in the end, i think it can be a good compromise, at the end the time overhead it would add (today or in the future I think) is very little or nothing when training the model (kernel and memory management and movement are what affect the performance the most in reality

StijnWoestenborghs:
Alright so first real usecase of References:
- TENSORS and GRADS are **not** globals anymore but back to a member of `model.parameters`
- When `model.parameters` is being changed in other parts of the code, it is being passed through as a **mutable reference**
- Also note the API `nn.optim.Adam[g](Reference(model.parameters), ... )`  In which case the optimizer is storing a mutable reference to the model parameters so it can update them.

basalt/nn/optim.mojo
```diff
@@ -22,8 +22,12 @@ fn get_trainable_parameters(g: Graph) -> List[Symbol]:
 
 struct Adam[
     g: Graph,
+    mutability: __mlir_type.i1,
+    lifetime: AnyLifetime[mutability].type,
     trainable_parameters: List[Symbol] = get_trainable_parameters(g),
 ]:
+    var parameters: Reference[Parameters, mutability, lifetime]
```
andresnowak, line 29: Okay so it is possible right now to have references in a struct (i just didn't know how it seems)

basalt/autograd/ops/dynamics.mojo
```diff
@@ -32,7 +32,13 @@ struct CONCAT:
...
     fn forward[
         attributes: AttributeVector,
-    ](inputs: List[Symbol], outputs: List[Symbol]):
+        mutability: __mlir_type.i1,
+        lifetime: AnyLifetime[mutability].type,
+    ](
+        inputs: List[Symbol],
+        outputs: List[Symbol],
+        parameters: Reference[Parameters, mutability, lifetime],
+    ):
```
andresnowak, line 41: Is it necessary to use reference here?, isn't it the same to use borrowed?

StijnWoestenborghs: Damn, yes ofcourse..., making thing unnecessarily complicated here

StijnWoestenborghs: Fixed

andresnowak:
Okay, for me it seems now everything is okay.

andresnowak (approved):

PR #62, opened 2024-04-29
Create a simple onnx data loader

andresnowak:
This data loader only loads the data in order for now (It doesn't use the names of the parameters)

- Also another thing is that we can only load data, it isn't possible to load the model because it has to be at compile time and I think it isn't possible to run python code at comp time.
- And  if we can load using the names of the parameters we still need to take care of the order (like if there is something like conv2d will always have the name conv2d"n") we define the layers or name them manually correctly, because if we have something like multiple operators where there results are concatenated or just using and add operator receives as input two other operators results, then there can be different ways one can define the order of the operators

Added:
- [x] Load onnx model data without names
- [x] Add attributes to node json and to the onnx graph 
- [x] Add ability to export to onnx model
   - torch can't import onnx models, there is a library onnx2torch, but it doesn't support all operators (e.g. Reshape)
   - Also logic of many operators work differently in onnx, like reshape instead of the shape being an attribute it is an input (They changed because it can be a dynamic value like the result of an operator, i don't know what cases maybe the shape of the result of another operator, but it can still be an attribute no problem). other operators like reduce_max have the same thing were the axis is an input instead of an attribute''

And in the future I think it is possible to add support for safetensors, the format is very simple it seems [safetensors format](https://github.com/huggingface/safetensors?tab=readme-ov-file#format)

**And I think it would be better to use naming in the future, but one can have that if loading it via naming doesn't work (because the names are different) it can then try to load it via order of the declaration of the variables.**


StijnWoestenborghs:
Nice,
So as i get this right. If one was to assume the order of nodes (as defined in a basalt graph) matches the onnx file. We could load the parameters without the need for naming the Symbols. Is that right? 
(If we don't have some kind of an assumption like this, the only other way i can see is to name the tensors manually or, automatically holding on the a similar naming convention as onnx.)
Any opinion on the best way here?

andresnowak:
Yeah from what I understand if one names the parameters (trainable weights) in the same order one can load the model without a problem. But what is the best solution i think both I would say, having the ability to use naming and order I think

StijnWoestenborghs:
Okay for me this is pretty good. Everything I checked looks good to me. A couple of things I think we should keep in mind:
- I think typing is not handled completely in the optimal way. But since we also haven't really cared about datatypes in the API so far. Having type as an extra field in attributes is a good starting point. (but good data type handling is for sure a different PR)
- We still need take care about naming the symbols (different PR) I don't think comp time will increase as last time. We found out that the root cause of this was string manipulations (in uuid) which are very unoptimized. But just storing a string as bytes is not a problem.
- A question that remains is how to handle the operators that are have attributes as an input of the node. We'll have to address it to get yolov8 to work. But definitely OK with it being a seperate PR as well.

andresnowak:
So do you want to merge it, or is there still something you want from this pr?

PR #63, opened 2024-05-02
Reworked Testing

Benny-Nottonson:
This pull request includes a series of commits that refactor the code and splits the tests into two categories, Python and Mojo. Mojo tests don't rely on any Python code and instead use hardcoded test cases, Python tests sometime use Torch or Numpy to dynamically test data.


StijnWoestenborghs (approved):
I'm good with this :+1:

PR #64, opened 2024-05-02
Add Basalt Docs website

Benny-Nottonson:
Added a NextJS website that is dynamically generated from the output of `mojo docs .` on the Basalt package, this is an introductory PR, the website itself will be updated with time.


andresnowak:
Maybe its okay if it is like this, but I think this should be in another repository no?, like a docs repo or webpage repo. Being sincere thats how I thought it was being made I don't know why. If its to much work then no problem but I feel this shouldn't be in the repo of the library I think, because it is a webpage and docs (not only docs)

Benny-Nottonson:
> Maybe its okay if it is like this, but I think this should be in another repository no?, like a docs repo or webpage repo. Being sincere thats how I thought it was being made I don't know why. If its to much work then no problem but I feel this shouldn't be in the repo of the library I think, because it is a webpage and docs (not only docs)

Me and @StijnWoestenborghs talked about it and I believe came to the conclusion that it makes more sense being part of the Basalt package, but it would be very easy to move it to a separate repo. The idea with it being in the same package is potentially a GH workflow that runs `mojo docs basalt` on release and updates the website immediately, something that is MUCH harder if its in a separate repo, open to either option though.

andresnowak:
okay, I thought it wasn't super difficult to have the mojo docs basalt being run from another repo in the gh workflow, didn't know. If it isn't possible to do it via another repo with the gh workflow, then I think it is okay probably.

Benny-Nottonson:
> okay, I thought it wasn't super difficult to have the mojo docs basalt being run from another repo in the gh workflow, didn't know. If it isn't possible to do it via another repo with the gh workflow, then I think it is okay probably.

I honestly agree that it would ideally be in a separate repo, but in my experience the workflow setup for that is complicated (See https://stackoverflow.com/questions/64374179/how-to-push-to-another-repository-in-github-actions), also keeping it in the directory of the project folder __may__ allow for more documentation features later like seeing the actual code for a function in the web, etc, things only available if you have access to the source.

StijnWoestenborghs:
Tbh, I don't know the best/right way here. Looking at other big projects (numpy/pytroch) They seem to be having docs related to tutorials and API references inside the repo (as that is changing with the library). But not any code related to the webpage. Which makes sense to do it like that, I guess. That looks at least like something manageable.

Probably if you release docs with the same release tag as the library. You can setup a workflow in the docs repo that just git pulls the library on the specified tag, and then releases the webpage. That shouldn't be too difficult I think. If you want I can have a look at that.

Does this split fit your current setup?

Benny-Nottonson:
i'm good with that 👍🏻 i'll close this pr later today and open a new repo for it. I'll probably hold off on writing the workflow for a bit since I can always manually pull changes aswell

Benny-Nottonson:
Closed and moved to github.com/basalt-org/docs :)

PR #66, opened 2024-05-03
Simplify if statement and syntax

Benny-Nottonson:
I have simplified the if statement and removed the size field in the Attribute struct. This change improves the code readability and removes unnecessary code.

### DIFF
--- a/basalt/autograd/attributes.mojo
+++ b/basalt/autograd/attributes.mojo
@@ -83,7 +83,6 @@ struct Attribute(Stringable, CollectionElement):
     var data: Bytes[MAX_DATA_BYTES]
     var data_shape: StaticIntTuple[MAX_RANK]
     var type: AttributeType
-    var size: Int
 
     @always_inline("nodebug")
     fn __init__(inout self, name: String, value: String):
@@ -92,7 +91,6 @@ struct Attribute(Stringable, CollectionElement):
         self.data_shape = StaticIntTuple[MAX_RANK]()
         self.data_shape[0] = len(value)
         self.type = AttributeType.STRING
-        self.size = 1
 
     @always_inline("nodebug")
     fn __init__(inout self, name: String, value: TensorShape):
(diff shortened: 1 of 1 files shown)


andresnowak:
The size value is something necessary, to be able to represent the size the same way for tensorshape, statictuple and others if they are added (so as to be able to call them the same type, in this case INTS), but everything else is okay, feel stupid that I didn't know there was a is_floating_point call haha

basalt/autograd/attributes.mojo
```diff
@@ -83,7 +83,6 @@ struct Attribute(Stringable, CollectionElement):
     var data: Bytes[MAX_DATA_BYTES]
     var data_shape: StaticIntTuple[MAX_RANK]
     var type: AttributeType
-    var size: Int
```
andresnowak, line 86: This value is necessary, to be able to represent things like statictuple and tensorshape the same way (in the end they are all lists of a value, so they are the same)

Benny-Nottonson:
okay, I actually like this as it might help with simplifying the code. I'll give this another look when i'm home today.

Benny-Nottonson:
What about storing Size/Type/other MetaData in a seperate AttrributeMetaData struct as one thing that way we don't have wasted fields for certain data types?

Benny-Nottonson:
for example, Scalar we need the Type information stored, but size won't matter at all, Shape we need Size but type is Int, if we add say Tensor later, we need Type and Size, but Int we need neither

andresnowak:
Something like an attribute metadata i would say is to much I think, and the only thing needed really here is type and size, like we would never have a tensor attribute really that will always be an input. The only type of attributes one can have are ints, floats, bool, string and a list of ints, floats and strings. So tensor shape still has a size because it is a list. Like even the attribute how it works right now was designed because of the problems we have with compile time. Like in reality it could just have been a list[variant], with a name and that's all.
But yeah having the size for float, int, bool and string is something extra one could say, but its just one line really, and metadata would be more extra I think, the only metadata is type and size and that can be included in the Attribute struct

Benny-Nottonson:
KK, ill give it a second pass right now

Benny-Nottonson:
Ready for another check @andresnowak 👍

andresnowak (approved):
I think it is okay now

PR #67, opened 2024-05-04
Use String `split` in MNIST dataset

fnands:
This MR implements a TODO that was in the MNIST dataset. 

This somewhat simplifies the logic of parsing the CSV file containing the MNIST data. 

I ran `mojo format` over the file, which changed the formatting in the BostonHousing dataset as well. 

I'll also go fix the TODO in the BostonHousing dataset once this one is in.

### DIFF
--- a/basalt/utils/datasets.mojo
+++ b/basalt/utils/datasets.mojo
@@ -59,7 +59,7 @@ struct MNIST:
 
     fn __init__(inout self, file_path: String) raises:
         var s = read_file(file_path)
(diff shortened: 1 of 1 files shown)


StijnWoestenborghs:
Nice, that reduces the time of loading the dataset as well. 
All good to me

StijnWoestenborghs (approved):

StijnWoestenborghs:
I oversaw something here. I think the data is not being loaded correctly now. 
You can always try to plot the image by uncommenting
https://github.com/basalt-org/basalt/blob/main/examples/mnist.mojo#L74

PR #69, opened 2024-05-04
Fix bug and improve code logic

Benny-Nottonson:
This pull request includes a fix for a bug and some improvements to the code logic. The bug fix addresses an issue related to converting a floating-point number to a sequence of bytes in IEEE 754 format. Instead of converting to F64 and storing Bytes from there, now we convert to UIntN and store those Bytes with bitshifting and bitcast.

### DIFF
--- a/basalt/autograd/attributes.mojo
+++ b/basalt/autograd/attributes.mojo
@@ -1,7 +1,7 @@
 from collections import Optional, OptionalReg
 
 from basalt.nn.tensor import Tensor, TensorShape, MAX_RANK
-from basalt.utils.bytes import Bytes, f64_to_bytes, bytes_to_f64
+from basalt.utils.bytes import Bytes, scalar_to_bytes, bytes_to_scalar
 
 
 alias MAX_ATTRS = 10
@@ -9,23 +9,17 @@ alias MAX_NAME_CHARS = 16
 alias MAX_DATA_BYTES = 32
 
 
-@value
 @register_passable("trivial")
 struct AttributeType(Stringable):
-    """
-    Attributes type.
-    """
-
-    alias INT = AttributeType(0, "INT")
-    alias FLOAT = AttributeType(1, "FLOAT")
-    alias STRING = AttributeType(2, "STRING")
-    alias INTS = AttributeType(3, "INTS")
-    alias FLOATS = AttributeType(4, "FLOATS")
-    alias STRINGS = AttributeType(5, "STRINGS")
-    alias BOOL = AttributeType(6, "BOOL")
+    alias BOOL = AttributeType(0, "BOOL")
+    alias INT = AttributeType(1, "INT")
+    alias FLOAT = AttributeType(2, "FLOAT")
+    alias STRING = AttributeType(3, "STRING")
+    alias INTS = AttributeType(4, "INTS")
+    alias FLOATS = AttributeType(5, "FLOATS")
 
     var id: UInt8
-    var name: Bytes[16]  # StringLiteral
+    var name: Bytes[16]
 
     fn __init__(inout self, id: UInt8, name: String):
         self.id = id
@@ -109,8 +103,6 @@ struct Attribute(Stringable, CollectionElement):
 
         for i in range(self.size):
             self.data_shape[i] = value._shape[i]
-        self.type = AttributeType.INTS
-        self.size = value.rank()
 
     @always_inline("nodebug")
     fn __init__[N: Int](inout self, name: String, value: StaticIntTuple[N]):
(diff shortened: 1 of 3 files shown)


StijnWoestenborghs:
Not sure about this
```
    fn __init__(inout self, name: String, value: FloatLiteral):
        self.__init__(name, Float32(value))
```
Maybe it is okay, but there is a potential pitfall. It looks like you cant switch between types now, or implicitly cast. 
The root cause is creating a scalar of f32 and getting it as a f64 or visa versa is not working now. 

The op in CLIP for example is getting the attribute as a scalar[dtype] (using the global dtype f32)
`min_attr.value().to_scalar[dtype]()`  
So that is now why it only works with casting all floatliterals to f32. Meaning that the user now only may input a f32 value and not a f64 or it will break.
So i think somehow that floatliteral is now constraining to have a specified input type. But maybe that is acceptable. I would say it is just something you can easily fall into and hard to find the issue.

Benny-Nottonson:
@StijnWoestenborghs in that case it may be better to store FloatLiteral as dtype (the global) if it is float point, otherwise just F64? let me know what you think

Benny-Nottonson:
it seems like more of an issue with API/FloatLiteral than an issue with attribute since it is properly storing the values for Scalar types

Benny-Nottonson:
I have an idea let me try something

Benny-Nottonson:
The only alternative I can think of it casting all lower sized types to the widest size then storing them that way, but that leads to other difficulties.

Benny-Nottonson:
@StijnWoestenborghs ready for review again 👍

andresnowak:
But isn't this adding a problem, where now a person can't add a negative value?, the scalar_to_bytes function is converting it to a uint no?

Benny-Nottonson:
> But isn't this adding a problem, where now a person can't add a negative value and instead would get an incorrect value?, the scalar_to_bytes function is converting the value to a uint first no (that is unexpected behavior)?. If one wants to fix the problem for big values, maybe an idea would be to have the f64_to_bytes conversion when the type passed is a float, and another one for ints when the value is an int, and that way we also preserve precision for float64 and can also have big values.

I dont think so, its not actually casting, its bitcasting, the only reason is to keep bitshifting logic the same for float and int, it should still work for negative numbers since it is bitcasting back at the end

Benny-Nottonson:
@andresnowak Essentially this changes the flow to 
```python
value: DType -> bitcasted to UInt_N (Where N is the number of Bytes) -> stored as bytes
or 
bytes -> UInt_N -> bitcast to DType
```

andresnowak:
hmmm okay I understand thats true it is a bitcast not a cast.

StijnWoestenborghs:
Okay I think i'm fine with this behaviour. It's a little hacky with using data_shape to know if it was created with a floatliteral. But that is the nature of Attributes i would say. There probably is an argument for using and extending AttributeType for this. But I guess its OK to leave it like this.
So now the only way it can fail would be by explicitly defining lets say a f64 scalar in the CLIP input. While your data is f32.

StijnWoestenborghs (approved):

Benny-Nottonson:
> Okay I think i'm fine with this behaviour. It's a little hacky with using data_shape to know if it was created with a floatliteral. But that is the nature of Attributes i would say. There probably is an argument for using and extending AttributeType for this. But I guess its OK to leave it like this. So now the only way it can fail would be by explicitly defining lets say a f64 scalar in the CLIP input. While your data is f32.

The other option is to cast all types up to UInt64, Float64, or Int64 depending on their type then casting back down to get it out, that way no issues with different types?

andresnowak:
The other thing is the conversion no? if one person passes let's say a float32 value, but when doing the to_scalar[float64] it also gives an incorrect value, that would be the other problem i think

Benny-Nottonson:
> The other thing is the conversion no? if one person passes let's say a float32 value, but when doing the to_scalar[float64] it also gives an incorrect value, that would be the other problem i think

Yeah I think this is what he is talking about, if either of you wants to give the casting up/down a shot it seems like it would work, otherwise I will try in a bit. Lets hold off on merging til we work that out.

StijnWoestenborghs:
Yeah OK I approved it but maybe it is not a bad idea to actually cast all to f64 and int64. That is how it used to work as well. And then there is no issue with implicit casts between 32 and 64

andresnowak:
Okay so you are saying casting it up when receiving them in the init function and then when doing the to_scalar one first does the conversion in the float64 or int64 and then casts it to float32 or whatever the user wants?

StijnWoestenborghs:
Yes exactly, that was how we did it before as well

Benny-Nottonson:
Almost done, testing now

Benny-Nottonson:
@StijnWoestenborghs @andresnowak 👍

StijnWoestenborghs:
I can feel the irony in that commit comment. It's good for me now :smile:

andresnowak (approved):
Also good for me

Benny-Nottonson:
I have more changes hold on 💀

Benny-Nottonson:
@StijnWoestenborghs @andresnowak Last one lol 🤞

PR #70, opened 2024-05-04
Formatting

Benny-Nottonson:
Ran `mojo format .`

### DIFF
--- a/basalt/autograd/graph.mojo
+++ b/basalt/autograd/graph.mojo
@@ -57,9 +57,13 @@ struct Graph:
         self.symbol_count += 1
         return scalar_id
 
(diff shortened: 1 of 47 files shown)


fnands:
That's a lot of changes! Have you tried seeing what it looks like with `-l 120`? 

Pytorch uses [120](https://github.com/pytorch/pytorch/blob/76a26a885d17fc6063c60207dd3195a1623f86df/pyproject.toml#L43) and IMO for a language like python where indentation is important 80 is just too restrictive.

Benny-Nottonson:
Closed due to needed internal discussion, line length is a big issue but there are other problems with the current formatter, will reopen when we have more style direction

PR #71, opened 2024-05-04
Add split to mnist

fnands:
Hi, trying this again (apologies for the last one). 

This time I confirmed the loss stayed the same, and visualized the images.  

I formatted with `mojo format -l 119`, but let me know if you prefer 80.

### DIFF
--- a/basalt/utils/datasets.mojo
+++ b/basalt/utils/datasets.mojo
@@ -76,11 +76,10 @@ struct MNIST:
 
         # Normalize data
         alias nelts = simdwidthof[dtype]()
-        var res = Tensor[dtype](self.data.shape())
 
         @parameter
         fn vecdiv[nelts: Int](idx: Int):
-            res.store[nelts](idx, div(self.data.load[nelts](idx), 255.0))
+            self.data.store[nelts](idx, div(self.data.load[nelts](idx), 255.0))
 
         vectorize[vecdiv, nelts](self.data.num_elements())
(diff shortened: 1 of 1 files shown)


Benny-Nottonson (approved):
There are some other changes im seeing in this file that need work, but they aren't relevant to this PR so I will save them for a separate branch, other than that it looks good to me.

StijnWoestenborghs:
Nice, those changes are good for me. This does decreses loading quite a lot actually.
Did see something else we were doing wrong from the beginning though. We were never using the normalized data. https://github.com/fnands/basalt/blob/add_split_to_mnist/basalt/utils/datasets.mojo#L77-L85
Could you fix that in this one? Something like:
```
from basalt import dtype, nelts
...
# Normalize data
@parameter
fn vecdiv[nelts: Int](idx: Int):
    self.data.store[nelts](idx, div(self.data.load[nelts](idx), 255.0))
vectorize[vecdiv, nelts](self.data.num_elements())
```

StijnWoestenborghs:
Thanks!

fnands:
I still wanted to point out that the loss is a bit higher in this case `0.000287` vs  `0.00093`

StijnWoestenborghs:
Yeah I saw that as well. It is in any case insignificantly low. And comparable with the pytorch implementation. Its imo more important to have normalization in both the pytorch and basalt example.

PR #73, opened 2024-05-05
Add split housing

fnands:
Hi, 

Similar to what I did for MNIST, here using the stdlib string splitting to parse the housing data. 

This time I checked that the loss is the same, but please just double check the logic.

### DIFF
--- a/basalt/utils/datasets.mojo
+++ b/basalt/utils/datasets.mojo
@@ -87,30 +87,10 @@ fn read_file(file_path: String) raises -> String:
     return s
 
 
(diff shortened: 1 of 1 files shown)


StijnWoestenborghs (commented):
Thank a lot. The change look good to me.
Noticed that the functions`find_nth` and `num_lines` further in the file are now unused. Merging once you they are deleted :+1:

fnands:
Good catch! 

Removed

StijnWoestenborghs (approved):

PR #74, opened 2024-05-05
Spring Cleaning

Benny-Nottonson:
Just some minor changes to some utils to clean up old code and get a little bit more performance :)

### DIFF
--- a/basalt/nn/tensor.mojo
+++ b/basalt/nn/tensor.mojo
@@ -47,6 +47,13 @@ struct TensorShape(Stringable):
         self._shape = shape
 
     @always_inline("nodebug")
+    fn __init__(inout self, owned shape: _TensorShape):
+        self._rank = shape.rank()
+        self._shape = StaticIntTuple[MAX_RANK]()
+        for i in range(min(self._rank, MAX_RANK)):
+            self._shape[i] = shape[i]
+
+    @always_inline("nodebug")
     fn __getitem__(self, index: Int) -> Int:
         return self._shape[index if index >= 0 else self._rank + index]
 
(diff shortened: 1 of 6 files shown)


Benny-Nottonson (commented):
I think using `[Type: DType]` in a function signature is definitely best practice if we have a global variable named dtype and we aren't asserting that `dtype == Type` although generally it is implied. With values I agree that we can use snake_case even for parameters, Types is a different story imho.

StijnWoestenborghs:
Also note that it should be confused with a type parameter. Connected to a trait. 
```fn type parameter	fn do_it[Action: Actionable](action: Action)	PascalCase```
Which I agree should be capital

**Comment by StijnWoestenborghs** [comment]
I'm very ok with the changes though. Nice cleanup on the perf utils. That made it much more readable.
Did you say you implemented that proposed init method in Tensor here as well by @fnands? `__init__(self, owned existing: MojoTensor)` I also think that is a good idea (maybe you forgot to push it)

This also made me realize again we have that backward SUB operation in MNIST that takes 87% of the total backward time. Probably caused by de-broadcasting. We should have a look at that at some point.
```14   | SUB            | 0.2663              | 87.453 % ```

**Comment by Benny-Nottonson** [comment]
I did forget to commit, good catch

File: basalt/nn/tensor.mojo [hunk 1]
```diff
@@ -126,6 +133,11 @@ struct Tensor[dtype: DType](Stringable, Movable, CollectionElement):
         self._data = data
         self._shape = shape
 
+    @always_inline("nodebug")
+    fn __init__(inout self, owned tensor: _Tensor[dtype]):
```
fnands, line 137: 🥳

StijnWoestenborghs (approved):

PR #76, opened 2024-05-08
Add CIFAR10 dataset, dataloader and training scripts

fnands:
This PR adds the code required to train a CNN on the CIFAR10 dataset. 

I adapted the Dataloader already there to one that looks more like a Pytorch dataloader. 

I would make it generic, but at this moment Mojo doesn't seem to support generic assignment? 
I.e. I created a BaseDataset, so that the DataLoader can possibly call this, but I get a compiler error: 
`TODO: dynamic traits not supported yet, please use a compile time generic instead of 'BaseDataset'`

Maybe one of you knows how to make the compile time generics work here? 

I haven't fully checked how well the model trains (add accuracy?), but the loss is about the same for PyTorch and Basalt. 

The BasaltTensor `__init__(inout self, owned tensor: _Tensor[dtype]):` results in some kind of double free making the program segfault, so it's just a simple copy for the moment.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "mimage"]
+	path = mimage
+	url = https://github.com/fnands/mimage.git
--- a/basalt/autograd/attributes.mojo
+++ b/basalt/autograd/attributes.mojo
@@ -19,11 +19,11 @@ struct AttributeType(Stringable):
     alias FLOATS = AttributeType(5, "FLOATS")
 
     var id: UInt8
-    var name: Bytes[16]
+    var name: Bytes[MAX_NAME_CHARS]
 
     fn __init__(inout self, id: UInt8, name: String):
         self.id = id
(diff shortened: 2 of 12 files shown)


fnands:
I added mimage as a submodule, but can also just add the `.mojopkg` file to the repo. 
Until such time as Mojo get's a packaging system I think those are the two options.

StijnWoestenborghs:
Thanks a lot for this, I looked at it and the example is solid. The thing I'm struggling with the most is it introducing a dependency to "all of basalt" while it is used for an "example using basalt". But the thing is, we have YoloV8 coming up, so we'll need the ability to read images for that as well. Maybe the right place for this (and all our examples) is a dedicated `basalt-examples` repo, as also the number of examples are growing as well.

Right now I'm thinking to leave this open and postpone a decision on this, to see how the other computer vision example turns out. Also the fact that there is no official package manager makes things harder.

PR #79, opened 2024-05-12
Add init and get methods for slice

Benny-Nottonson:
This pull request adds the `__init__` and `to_slice` methods to the `Attribute` struct. The `__init__` method allows for the initialization of an attribute with optional integer values, while the `to_slice` method converts the attribute data into a static tuple of optional integer values. This addition enhances the functionality of the `Attribute` struct and improves code readability and flexibility.


StijnWoestenborghs:
@Benny-Nottonson As Slice took a turn in how the attributes are used, we can probably just close this one. As we don't the functionality anymore

PR #107, opened 2024-05-16
Implement leaky_relu

FrostyTheSouthernSnowman:
The README mentioned that more activation functions are on the road map. I have gotten started. As of publishing, I only added Threshold, but I plan to go down the list from the (PyTorch Docs)[https://pytorch.org/docs/stable/nn.functional.html#non-linear-activation-functions]. Since I'm just going down the list, I'll do HardTanh next, unless someone has one they want written first. I am willing to write documentation if that is needed. Otherwise, I am trying to keep everything as close to the PyTorch docs as possible. If anybody thinks I should follow Tensoflow, Jax, or something else's docs instead, I am totally up for it as well. Here's the current TODO list for what I plan to implement first:

- [ ] threshold
- [ ] hardtanh
- [ ] hardswish
- [ ] selu
- [ ] celu
- [x] leaky_relu
- [ ] gelu
- [ ] softsign
- [ ] softplus
- [ ] softmin
- [ ] hardsigmoid
- [ ] mish
- [ ] normalize

Once I get all of those done, I'll do the rest of the ones listed on the PyTorch docs, assuming my schedule allows for it. I am more than happy to switch out anything on the list.

Also, I've noticed that the current AttributeVector system lead to a lot of type conversions, which I feel could hinder performance. I don't totally plan to look into it right now, although I could if it were necessary. 

There also seem to be a lot of overloads in the test file. Once I implement a few more activation functions, I'd like to see if I can split it into just a few basic categories that can be widely used across all the different activation functions.

### DIFF
--- a/basalt/autograd/ops/mlops.mojo
+++ b/basalt/autograd/ops/mlops.mojo
@@ -7,65 +7,6 @@ from basalt.utils.tensorutils import elwise_transform
 from basalt.autograd.attributes import Attribute, AttributeVector
 
 
-struct THRESHOLD:
-    @staticmethod
-    fn result_shape(t1_shape: TensorShape) -> TensorShape:
-        return t1_shape
-
-    @staticmethod
-    fn forward[
-        t1_shape: TensorShape,
-        attributes: AttributeVector,
-    ](inout res: Tensor[dtype], t1: Tensor[dtype]):
-        """Forward pass for threshold."""
-
-        alias THRESHOLD: Scalar[dtype] = attributes[
-            "threshold"
-        ].value().to_scalar[dtype]()
-
-        alias VALUE: Scalar[dtype] = attributes["value"].value().to_scalar[
-            dtype
-        ]()
-
-        @always_inline
-        fn threshold[
-            type: DType, simd_width: Int
-        ](x: SIMD[type, simd_width]) -> SIMD[type, simd_width]:
-            alias casted_threshold = THRESHOLD.cast[type]()
-            alias casted_value = VALUE.cast[type]()
-
-            return (x > x.splat(casted_threshold)).select[type](x, casted_value)
-
-        elwise_transform[threshold](res, t1)
-
-    @staticmethod
-    fn backward[
-        ug_shape: TensorShape,
-        t1_shape: TensorShape,
-        attributes: AttributeVector,
-    ](ug: Tensor[dtype], t1: Tensor[dtype]) -> Tensor[dtype]:
-        """Backward pass for threshold."""
-        alias THRESHOLD: Scalar[dtype] = attributes[
-            "threshold"
-        ].value().to_scalar[dtype]()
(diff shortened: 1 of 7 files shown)


andresnowak:
For the activation functions it would be in tests/mojo/test_mlops and tests/python/test_mlops_torch. And I wouldn't say there are activation function to prioritize, but i would say the only necessary for now would be leaky_relu, gelu and selu (from the ones mentioned i think those are the most used). Because maybe we have to think about how to divide some parts of the code base, so some cleanups will be necessary. So I think with only those function would be enough I think.

FrostyTheSouthernSnowman:
Ok. Will work on leaky_relu and selu, and I'll do gelu last.

FrostyTheSouthernSnowman:
Also, I was looking through the code and I realized that there are tests in `test_mlops.mojo` for the activation functions that I forgot to write. Will write those for `threshold`, `hard_tanh`, and `leaky_relu`, as well as the tests in `test_activations`

andresnowak:
There is also tests in tests/python/test_mlops_torch.mojo. Those are the three places test_activations, test_mlops and test_mlops_torch

FrostyTheSouthernSnowman:
The torch compatibility tests for threshold and hard tanh show there to be some bugs. Are they important enough to be worth debugging or should I just continue to gelu and selu and just delete all the threshold and hardtanh code?

andresnowak:
If there are errors when comparing with the torch version yeah they should be fixed. But if you want yeah you can delete them and work with gelu, selu or you can fix the hard thanh and threshold.

FrostyTheSouthernSnowman:
Ok. Haven't ever heard of them being used anyways. I'll just get rid of that code and work on the gelu and selu

FrostyTheSouthernSnowman:
Is there an OP for elementwise min or max? Like what is denoted mathematically by say min(0, x)? Would be useful in a few places. And could simplify some implementations.

andresnowak:
yeah in autograd/ops/basics there is the max op, or in utils/tensorutils there is the reduce op

andresnowak:
that is part of mojo, there is already a max op in mojo in the math module

FrostyTheSouthernSnowman:
Thanks! I'll use that then

FrostyTheSouthernSnowman:
Unfortunately I'm busier now than anticipated and I won't be able to complete the other activation functions. I do have full testing however for leaky_relu and I'd like to at least merge that in if possible.

andresnowak (approved):
Seems good

andresnowak:
And also thank you for the contribution

FrostyTheSouthernSnowman:
Oh thank you! No worries. Sorry I couldn’t finish what I hoped to.On Jun 28, 2024, at 11:32 PM, Andres Nowak ***@***.***> wrote:﻿
And also thank you for the contribution

PR #112, opened 2024-05-18
Added support for KAN (WIP)

Benny-Nottonson:
Not working yet, graph renders but operation causes a segfault at runtime.

### DIFF
--- a/basalt/autograd/graph.mojo
+++ b/basalt/autograd/graph.mojo
@@ -125,6 +125,22 @@ struct Graph:
         self.nodes.append(Node(OP.CONCAT, inputs, List[Symbol](res), attributes))
         return res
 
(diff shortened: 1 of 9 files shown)


andresnowak:
ONe of the problems is in your reshapes you are using -1 (we still don't have support for that). Instead they should be something like this.

```py
g.op(
        OP.RESHAPE,
        summed_two,
        attributes=AttributeVector(
            Attribute("shape", TensorShape(summed_two.shape.num_elements() // output_dim, output_dim))
        ),
    )
```

andresnowak:
Also the dot operators are doing calculations with incorrect shapes. Like: g.op(OP.DOT, acos, arange) is doing the calcualtion 32x1x1 @ 9, it should be @ 1x1x9  I think. There are many operations using wrong shapes from what i'm seeing.
And Also we still don't have batch_matmul

PR #115, opened 2024-05-20
Add MAE loss function

Benny-Nottonson:
Added an implementation of MAE loss, this also adds the ABS operator (Inplace absolute value) and its associated derivative. 

Signed-off-by: benny-nottonson <bennynottonson@gmail.com>


Benny-Nottonson (commented):
This needs tests before it can be fully merged

PR #116, opened 2024-05-21
Changes to Onnx loader

andresnowak:
**Fix onnx importing (problem not considering onnx considers in many operators attributes as inputs tensors)**

Some attributes in some operators of Onnx are considered as inputs instead of attributes. What majority do is create the attributes as "constant" (special type onnx node) type nodes, so in this pull request I changed the creation from initializers to constants. And because of this it seems we don't need to care about values being initializers instead for this special cases, e.g the yolov8 Onnx model uses constants, the only initializers are the weights and biases for the convolutions.

**Extra**: Also fixed an error where static tuple was using data instead of data shape, changed this so both tensorshape and staticInttuple can be interpreted as the same type (at the end of the day they are just lists values).

**Extra not important**: Also nightly is adding the parameter decorator for for loop and it seems it doesn't unroll the loop, so now we are able to change in the forward and backward function from models from unrolled to this parameter for loop (Because we don't want to unroll this loop, A graph can get very big and an unroll here doesn't really help really, it just makes compile times slower, or at least thats what I think) [parameter for loop](https://github.com/modularml/mojo/blob/nightly/docs/changelog.md#%EF%B8%8F-new)

_And also a recursive function that uses a parameter to determine when to end gets unrolled_, because of this it isn't a recursive function anymore so you can even make the function always_inline (I have tested this), so you can create nested for loops dynamically (And the speeds are exactly the same as a nested for loop, or when you just one for loop it is also the same speed, and it is faster than a recursive function that doesn't use a parameter value.)

### DIFF
--- a/basalt/autograd/attributes.mojo
+++ b/basalt/autograd/attributes.mojo
@@ -115,7 +115,7 @@ struct Attribute(Stringable, CollectionElement):
         self.size = N
 
         for i in range(self.size):
-            self.data[i] = value[i]
+            self.data_shape[i] = value[i]
 
     @always_inline("nodebug")
     fn __init__[dtype: DType](inout self, name: String, value: Scalar[dtype]):
@@ -156,7 +156,7 @@ struct Attribute(Stringable, CollectionElement):
         var result = StaticIntTuple[N]()
 
         for i in range(N):
-            result[i] = int(self.data[i])
+            result[i] = int(self.data_shape[i])
 
         return result
 
--- a/basalt/utils/onnx_utils.mojo
+++ b/basalt/utils/onnx_utils.mojo
@@ -6,32 +6,12 @@ from basalt.nn.model import Parameters
 from basalt.nn.tensor import Tensor, TensorShape
 from basalt.autograd.attributes import Attribute, AttributeType
 from basalt.autograd.ops import OP
+from basalt.autograd.graph import Node
+
+from .tensor_creation_utils import to_numpy, copy_np_data
 
 # NOTE: Maybe we could create our own model representation and from there convert to onnx or others (well we already have it in reallity)
 # NOTE: Torch doesn't import onnx, need onnx2torch and it doesn't support operators like reshape?
-
-fn to_numpy(tensor: Tensor) raises -> PythonObject:
-    var np = Python.import_module("numpy")
-
-    np.set_printoptions(4)
-    var rank = tensor.rank()
-    var pyarray: PythonObject = np.array([0])
-
-    if rank == 1:
-        pyarray = np.empty((tensor.dim(0)))
-    elif rank == 2:
-        pyarray = np.empty((tensor.dim(0), tensor.dim(1)))
-    elif rank == 3:
-        pyarray = np.empty((tensor.dim(0), tensor.dim(1), tensor.dim(2)))
-    elif rank == 4:
-        pyarray = np.empty((tensor.dim(0), tensor.dim(1), tensor.dim(2), tensor.dim(3)))
-    else:
-        print("Error: rank not supported: ", rank)
-
-    for i in range(tensor.num_elements()):
-        pyarray.itemset((i), tensor[i])
-
-    return pyarray
 
 
 fn make_onnx_attribute(op: OP, attr: Attribute) raises -> PythonObject:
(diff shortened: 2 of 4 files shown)


StijnWoestenborghs:
Totally agree in creating constant inputs for these cases. (Actually I was about to do something similar in upsampling)
Those comments are super interesting. 
- That parameter foor loop will be super convenient. Totally ready have the unroll gone in model
- Now that you are saying that about the recursive functions, I might rethink the product itertools actually. That is reassuring for the ndim slice for sure. I'm interested, how did you test this? Solely on speed or did you go deeper?

andresnowak:
LIke this

```py
from benchmark import benchmark
from autotune import cost_of


# This function is not recursive in the end thanks to the parameter value to denoting the end of the recursion at comp time (so the recursion probably gets unrolled and instead we just get nestd for loops?). And because of this the function can be always inline.
@always_inline
fn recursive_for[pos: Int, end: Int = 1](val: Int, size: Int, inout sum: Int):
    @parameter
    if pos == end:
        sum += val
        return
    else:
        for i in range(size):

            @parameter
            if pos == end - 1:
                recursive_for[pos + 1, end](val + i, size, sum)
            else:
                recursive_for[pos + 1, end](val + i * size, size, sum)


fn recursive_for_bad(pos: Int, end: Int, val: Int, size: Int, inout sum: Int):
    if pos == end:
        sum += val
        return
    else:
        for i in range(size):
            if pos == end - 1:
                recursive_for_bad(pos + 1, end, val + i, size, sum)
            else:
                recursive_for_bad(pos + 1, end, val + i * size, size, sum)


fn normal_for(size: Int, inout sum: Int):
    for i in range(size):
        sum += i


fn normal_for_nested(size: Int, inout sum: Int):
    for i in range(size):
        for j in range(size):
            sum += i * size + j


alias recursive_for_type = fn (Int, Int, inout Int, /) -> None
alias normal_for_type = fn (Int, inout Int, /) -> None
alias normal_for_nested_type = fn (Int, inout Int, /) -> None
alias recursive_for_bad_type = fn (Int, Int, Int, Int, inout Int, /) -> None


fn main():
    print("yeah")

    var h = 0

    @parameter
    fn wrapper():
        h = 0
        recursive_for[0](0, 1_000_000_000, h)

    @parameter
    fn wrapper2():
        h = 0
        normal_for(1_000_000_000, h)

    var r_for = benchmark.run[wrapper]()
    print(h)
    var n_for = benchmark.run[wrapper2]()
    print(h)

    r_for.print()
    n_for.print()

    print(cost_of[normal_for_type, normal_for]())
    print(cost_of[recursive_for_type, recursive_for[0]]())
    print()
    print(cost_of[normal_for_nested_type, normal_for_nested]())
    print(cost_of[recursive_for_type, recursive_for[0, 2]]())
    print()
    print(cost_of[recursive_for_bad_type, recursive_for_bad]())
    print()

    @parameter
    fn wrapper3():
        h = 0
        recursive_for[0, 2](0, 10_000, h)

    @parameter
    fn wrapper4():
        h = 0
        normal_for_nested(10_000, h)

    @parameter
    fn wrapper5():
        h = 0
        recursive_for_bad(0, 2, 0, 10_000, h)

    var r_for_nested = benchmark.run[wrapper3]()
    print(h)
    var n_for_nested = benchmark.run[wrapper4]()
    print(h)
    var r_for_bad = benchmark.run[wrapper5]()
    print(h)

    r_for_nested.print()
    n_for_nested.print()
    r_for_bad.print()

    benchmark.keep(h)
```

StijnWoestenborghs:
I'm aware Yolov8 also has those examples where some attributes are considered as node inputs. I'm thinking that that is a pretty good test for the fixes here. So for me it's good to merge this as preparation for it actually and overkill this with unittests. What do you think?

andresnowak:
For me is also okay, I also think unitests maybe they should be necessary but I think for now lets use the yolov8 example as the test (it is a little bit overkill the tests at least right now for the onnx I think).

StijnWoestenborghs:
Nice, I didn't know about the existence of `cost_of` that is a SUPER convenient utility.

StijnWoestenborghs (approved):

andresnowak:
yep before it didnt work correctly but now it does. very cool

StijnWoestenborghs:
Definitely using that when looking at matmul again

PR #119, opened 2024-06-08
Start update to v24.4

andresnowak:
- [x] Fix mnist example
- [x] Fix test_backward inf error (maybe it is because of the change from neginf to min_or_neg_inf). The problem was Did a fix for the pow gradient of the exponent when values in t1 are 0 (in the log op of the backward), because log of 0.0 is -inf (and I don't know if its correct it seems some just make it zero when value is inf sometimes, but I don't know what is the correct solution).
- [x] make a math_util (round, add, sub, mul, div)
- [x] Check collection reference return
- [x] remove @parameter fors
- [x] Remove some unnecesary always_inline (don't help in speed and add binary size and comp time, like inits) and because those structs are to big to be register passable really.
- [x] Make model use @parameter for instead of unroll (to see if compile times get faster). **For other parts of code for now we can leave the unroll, I think its better to wait until modular removes it, if they do.**
- [ ] use string literal instead of bytes for names (**Memory stuff seems now is starting to have a lot of less errors at comptime, it is now possible to use string literal instead of bytes in attributes, op, etc.. but i'm not going to change it for now because everything for now is register passable values (and we have to change that, of course it was done because there were errors but now it seems it is possible to start changing this), the amount of data we save in them in the first place is to much to be register passable, also modular seems to be doing some changes to let less things be register passable (because it shouldn't be possible in the first place).**)

Compile times don't seem to be faster, i'm removing some parameter fors like in bytes because it is not neccessary and just adds unrolls (comp time and binary size) and some inlines like init of some structs (doesn't add speed and increases binary sizes and compt times), but it doesn't seem that comp times get faster

Also maybe some of the slow speed in comp time is because of all the copies that happen because all our values are register passable (and register passable (at least trivial) always makes a copy)

To completely know the comp times, we need to test with yolov8, if it isn't faster then maybe a lot of copies and unroll happens and it is our fault or it is also because there can be more optimizations in the compiler. But i'm not sure if they are faster.

Also it literally seems everything is starting to work at comptime for memory, but getting lifetimes from a comptime value seems to still not be possible or there are some sharp edges. And also inline array can do setitem like statictuple

### DIFF
--- a/basalt/autograd/attributes.mojo
+++ b/basalt/autograd/attributes.mojo
@@ -1,4 +1,5 @@
 from collections import Optional, OptionalReg
+from utils.static_tuple import StaticTuple
 
 from basalt.nn.tensor import Tensor, TensorShape, MAX_RANK
 from basalt.utils.bytes import Bytes, scalar_to_bytes, bytes_to_scalar
@@ -47,7 +48,7 @@ struct AttributeVector(Sized, Stringable, CollectionElement):
 
     @always_inline("nodebug")
     fn __init__(inout self, *attributes: Attribute):
-        self.attributes = StaticTuple[Attribute, MAX_ATTRS]()
+        self.attributes = StaticTuple[Attribute, MAX_ATTRS](Attribute("", ""))
         self.size = len(attributes)
         for i in range(self.size):
             self.attributes[i] = attributes[i]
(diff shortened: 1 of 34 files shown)


StijnWoestenborghs:
Nice, thank a lot for these updates. 
I'll have to get used to the new updates on References 

Have you noticed that the performance of the MNIST example has increased? It is now as fast as pytorch for me 0.0

StijnWoestenborghs (approved):

andresnowak:
No, being sincere I didnt notice it, if it is awesome. Also for the compilation time I don't know if it is faster

StijnWoestenborghs:
Yeah don't known about compilation time as well, still takes ages

andresnowak:
i think that maybe a part is the attribute vector (when looking for each attribute)

PR #123, opened 2024-09-14
Update to Mojo 24.6

andresnowak:
## Update to 24.6
I wanted to at least update the library.

- [x] Error in backward tests. Fixed
- [x] compile error in mlops. Last missing test (mojo test) to fix for update
    - [x] The compile error is slice_axes in forward and backward, and slice in backward
    - [x] And also slice has an error in implementation. The error was in attributes to_list function
- [x] Error in MNIST, where the loss is going up. Fixed error was related to backward test
- [x] Python tests
- [x] Mnist Load

All the library for the tests and examples were updated

Extra:
- IF you run in test_tensorutils only the test_accumulate_grad function there is a crash, seems like a mojo bug.
- Test_ops_torch takes a very long time to compile
- Test_mlops_torch also takes a also a very long time to compile


Benny-Nottonson:
@andresnowak what version is this supposed to work on?

andresnowak:
24.6, the newest one

PR #125, opened 2025-02-13
Added GELU activation function

Puchungualotsqui:
I've added GELU activation. Used in Bert and Gpt.


Puchungualotsqui:
I would like to add more activation functions. I am having a final project comparing the performance with Keras. Please, tell me if the implementation is correct. And is there the OP.WHERE or an equivalent?

